PR #42, opened 2017-02-24
Fix include/exclude pattern description in GUIDE.md

danthelion:
Include/Exclude pattern prefixes have been switched up in the documentation. It could cause some confusion. :)


gilbertchen:
Thank you for the fix.  This was a really bad typo.

PR #73, opened 2017-06-13
Fix typo

sdaros:
The file name should be "known_hosts"


gilbertchen:
Thank you for the fix!

sdaros:
No problem

PR #81, opened 2017-06-19
Move error parsing behind status code handling

chbmuc:
Otherwise request throttling won't work and you will get errors like this:

PUT https://api.onedrive.com/v1.0/drive/root:/dup/chunks/91xxx08:/content
Failed to upload the chunk 91xxx08: 503 Unexpected response

### DIFF
--- a/src/duplicacy_oneclient.go
+++ b/src/duplicacy_oneclient.go
@@ -128,12 +128,6 @@ func (client *OneDriveClient) call(url string, method string, input interface{},
             Error:  OneDriveError { Status: response.StatusCode },
         }
 
-        if err := json.NewDecoder(response.Body).Decode(errorResponse); err != nil {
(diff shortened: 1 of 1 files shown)


gilbertchen:
Thank you for fixing this.  Since the source code is released under the Fair Source 5 license, I need to ask you if you agree to transfer the copyright of your code in this PR to me? As a compensation, I will share with you and other contributors 20% to 30% percent of the revenue from the sales of the Fair Source licenses (and I will publish the sales number every month).

chbmuc:
As this is almost a trivial change, I will grant all copyrights to you. No need to share any revenue for this!

gilbertchen:
The change may look trivial but it is not -- it would have taken me a lot of time had another user reported the error first. I'll add you to the contributor list.

PR #110, opened 2017-08-12
Update README.md

clbn:
Fixes typo ("serivce" -> "service") and factual error (B2 is not the least expensive since Wasabi was added).


gilbertchen:
Thanks for the catch.  Sadly, B2 is not the fastest any more (it was the fastest because it was the only to support multithreading at that time).  I'll fix that as well.

PR #139, opened 2017-08-28
Fix typos: snpashot -> snapshot

countextreme:
No logic changes, just fixed typo in strings


gilbertchen:
Thank you for the fix!

Issue #134, opened 2017-08-27
Copy Performance


jt70471:
Hi, I'm backing up to a local directory, then using COPY to copy snapshots to the backblaze b2 cloud. If I have the local and remote destinations in sync, performing a local snapshot that produces a small number of new chunks.

When I then copy the local repository snapshots using:

```
duplicacy copy -to b2
```
it would appear that duplicacy goes through each chunk and prints a message that the chunk has been copied for every chunk. I know this isn't possible, because my upload speed isn't fast enough to actually perform the copy of all chunks.

I would have expected the copy to only copy the chunks that were new in the last snapshot. I.e. List all chunks in the "-from" snapshots and all chunks in the "-to" snapshots, which would then be used to only upload the new chunks to the remote repository, which should drastically speed up COPY processing.

Is that possible or can you explain why I'm seeing COPY produce messages that it's copied each and every chunk, even when I know the "-to" location already has copies of most of the chunks?

Thanks.

gilbertchen:
```
            if skipped {
                LOG_INFO("SNAPSHOT_COPY", "Chunk %s (%d/%d) exists in the destination", chunk.GetID(), chunkIndex, len(chunks))
            } else {
                LOG_INFO("SNAPSHOT_COPY", "Copied chunk %s (%d/%d)", chunk.GetID(), chunkIndex, len(chunks))
            }

```

It could be a bug that `skipped` isn't set correctly.  I'll take a close look.

jt70471:
Hi, I have reviewed duplicacy_backupmanager.go and for the life of me, I don't see anywhere that the skipped flag is being set or even how it's supposed to be passed to the uploadCompletion function.

If you could give me a few hints how you think that should occur, I'm happy to try to figure it out.

Thanks

gilbertchen:
I found out the place where the `skipped` flag should be set: https://github.com/gilbertchen/duplicacy/commit/3ae2de241e37f1b10e147a96d37ee75eba30276b

PR #154, opened 2017-09-03
Skip chunks to copy if already on destination for issue #134

jt70471:
Hi @gilbertchen, here's a pull request for the copy performance optimization. I have tested on macOS, which is only OS I can test with. The copy will skip any chunks already on the destination.

The resulting destination storage seems ok because a "check -stats -files" confirms all chunks exist for all snapshots.

Please take a look and let me know if you'd like changes to the code, otherwise please consider it for inclusion into duplicacy. Thanks!


gilbertchen:
This looks good.  Thank you for your contribution!

Issue #57, opened 2017-06-02
Ignoring keys in preferences


danielfaulknor:
I have the following in preferences:

```
        "no_save_password": false,
        "keys": {
            "password": "password",
            "ssh_keyfile": "/root/.ssh/id_rsa"
        }
```

But when I run duplicacy backup, it still prompts for all those details. Have I missed something?

Running the latest release from the releases page

gilbertchen:
If you add a non-empty password it should work:
```
        "no_save_password": false,
        "keys": {
            "password": "password",
            "ssh_password": "x",
            "ssh_keyfile": "/root/.ssh/id_rsa"
        }
```

The SFTP storage always tries password login first, so if you don't provide one in the preferences file it will ask for it.

danielfaulknor:
> I might have a crack at a PR to use key first if it's defined as I don't
really want to fill my logs with invalid password attempts :)

danielfaulknor:
Hmm, just tried it out, it's still prompting for a ssh key. Setting the environment variable yields the same result:

[root@dev01 www]# DUPLICACY_SSH_KEY_FILE="/root/.ssh/id_rsa"
[root@dev01 www]# duplicacy backup
Storage set to sftp://SERVER/backups
Enter the path of the private key file:^C
[root@dev01 www]#

gilbertchen:
The key name should be `ssh_key_file`.  This is a typo in the user guide.

seegras3:
I've got an similar issue, my preferences looks like this

```
"no_save_password": false,
        "keys": {
            "DUPLICACY_B2_ID": "xxxxxxxxx",
            "DUPLICACY_B2_KEY": "xxxxxxxx",
            "DUPLICACY_PASSWORD": "xxxxxxxxx"
        }
```

when doing the Backup like this it still asks me for the Storage Password

```
../duplicacy -d backup
Storage set to b2://test-999
Reading the environment variable DUPLICACY_B2_ID
Reading the environment variable DUPLICACY_B2_KEY
Reading the environment variable DUPLICACY_PASSWORD
Failed to get the value from the keyring: keyring: Password not found
Enter storage password:**********
```

If i change the Key `DUPLICACY_PASSWORD` into just `PASSWORD` it works as expected.

I dont know if this is a Bug but i guess so

gilbertchen:
[...]

PR #155, opened 2017-09-04
Use ssh key file first if we have it in preferences/environment

niknah:
If we have a sftp key file in the environment/preferences, then don't attempt a password login to avoid bad login errors.

https://github.com/gilbertchen/duplicacy/issues/57


gilbertchen:
Thanks for fixing it.

Issue #151, opened 2017-09-03
add (secondary storage) does not respect env password


robbat2:
The `add` command to add secondary storage does not seem to respect the environment variables `DUPLICACY_<STORAGENAME>_PASSWORD` as documented at https://github.com/gilbertchen/duplicacy/blob/master/GUIDE.md. 

Actual output:
```
/tmp/backup will be backed up to /tmp/storage-dir-1 with id snapshot-id
Enter storage password for /tmp/storage-dir-2:
```

Expected output:
```
/tmp/backup will be backed up to /tmp/storage-dir-1 with id snapshot-id
/tmp/backup will be backed up to /tmp/storage-dir-2 with id snapshot-id
Storage set to /tmp/storage-dir-1
No previous backup found
Indexing /tmp/backup
Packed a (0)
Backup for /tmp/backup at revision 1 completed
```

Testcase:
```
#!/bin/bash
STORAGE_1=/tmp/storage-dir-1
STORAGE_2=/tmp/storage-dir-2
MYDATA=/tmp/backup
rm -rf $STORAGE_1 $MYDATA
mkdir -p $STORAGE_1 $STORAGE_2 $MYDATA
cd $MYDATA
touch $MYDATA/a
export DUPLICACY_PASSWORD=foobar123
export DUPLICACY_disk2_PASSWORD=foobar123
duplicacy init -e snapshot-id $STORAGE_1
duplicacy add -e disk2 snapshot-id $STORAGE_2
duplicacy backup
```

niknah:
Try DUPLICACY_DISK2_PASSWORD

In duplicacy_utils.go it looks like it runs ToUpper()  on the environment name first.

robbat2:
Ok, so either the documentation needs to be updated to say the uppercase of the name, or it should take the exact case of the name as provided.

gilbertchen:
This is a doc bug.

PR #161, opened 2017-09-05
Fix doc bug for issue #151

jt70471:
Fix doc bug referenced in issue #151.


gilbertchen:
Thanks for fixing this.

PR #165, opened 2017-09-06
Use number of threads specified on copy command

jt70471:
Hi, when specifying the "-threads" command line option for the copy command, duplicacy encounters an out of range error. This patch passes the number of threads from the command line to the CreateStorage calls in copySnapshots.


gilbertchen:
Thanks for fixing it.  The -threads option only changes the number of uploading threads. The downloading is always sequential.

jkl1337:
I apologize. I filed this issue thinking I was running master. Almost soon after I clicked send I realized that system was running the release!

Issue #169, opened 2017-09-08
Upload rate limit not working for copy


jt70471:
Hi, looks like the upload rate limit is not working for the copy command. Looks like the wrong variable names used.

gilbertchen:
Fixed by https://github.com/gilbertchen/duplicacy/pull/170.

PR #170, opened 2017-09-08
fix upload/download rate for copy described in issue #169

jt70471:


gilbertchen:
Thank you for fixing it.

PR #173, opened 2017-09-09
change message when chunk is skipped at destination for copy

jt70471:
The message when a chunk is optimized and skipped at the destination because it's referenced in one or more destination snapshots is too similar to the message when the destination is checked and finds that the chunk physically exists.

This patch clarifies when a chunk is skipped vs physically found at the destination when an attempt to upload the chunk is performed.


gilbertchen:
A good change!

PR #180, opened 2017-09-11
Added backup --dry-run option.

niknah:
Maybe useful for other people who want to test a large backup but don't want to run it yet.

### DIFF
--- a/src/duplicacy_backupmanager.go
+++ b/src/duplicacy_backupmanager.go
@@ -1698,12 +1698,10 @@ func (manager *BackupManager) CopySnapshots(otherManager *BackupManager, snapsho
         if revisionMap[snapshot.ID][snapshot.Revision] == false {
           continue
         }
-        if !manager.config.dryRun {
-          otherManager.storage.CreateDirectory(0, fmt.Sprintf("snapshots/%s", snapshot.ID))
-          description, _ := snapshot.MarshalJSON()
-          path := fmt.Sprintf("snapshots/%s/%d", snapshot.ID, snapshot.Revision)
-          otherManager.SnapshotManager.UploadFile(path, path, description)
-        }
+        otherManager.storage.CreateDirectory(0, fmt.Sprintf("snapshots/%s", snapshot.ID))
+        description, _ := snapshot.MarshalJSON()
+        path := fmt.Sprintf("snapshots/%s/%d", snapshot.ID, snapshot.Revision)
+        otherManager.SnapshotManager.UploadFile(path, path, description)
         LOG_INFO("SNAPSHOT_COPY", "Copied snapshot %s at revision %d", snapshot.ID, snapshot.Revision)
     }
 
--- a/src/duplicacy_backupmanager.go
+++ b/src/duplicacy_backupmanager.go
@@ -36,6 +36,10 @@ type BackupManager struct {
 }
 
 
+func (manager *BackupManager) SetDryRun(dryRun bool) {
+    manager.config.dryRun = dryRun
+}
+
 
 // CreateBackupManager creates a backup manager using the specified 'storage'.  'snapshotID' is a unique id to
 // identify snapshots created for this repository.  'top' is the top directory of the repository.  'password' is the
@@ -630,7 +634,9 @@ func (manager *BackupManager) Backup(top string, quickMode bool, threads int, ta
     }
     skippedFiles = append(skippedFiles, fileReader.SkippedFiles...)
 
-    manager.SnapshotManager.CleanSnapshotCache(localSnapshot, nil)
+    if !manager.config.dryRun {
+        manager.SnapshotManager.CleanSnapshotCache(localSnapshot, nil)
+    }
     LOG_INFO("BACKUP_END", "Backup for %s at revision %d completed", top, localSnapshot.Revision)
 
     RunAtError = func() {}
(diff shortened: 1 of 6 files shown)


niknah:
Sorry, I missed that in the merge.
I've also noticed that in some of the storage drivers(sftp, file), it makes a directory in FindChunk.  So it may make an empty directory during a --dry-run

src/duplicacy_snapshotmanager.go
```diff
@@ -600,7 +600,9 @@ func (manager *SnapshotManager) ListAllFiles(storage Storage, top string) (allFi
             if len(files) > 1024 && !storage.IsFastListing() {
                 for i := 0; i < 256; i++ {
                     subdir := dir + fmt.Sprintf("%02x\n", i)
-                    manager.storage.CreateDirectory(0, subdir)
+                    if !manager.config.dryRun {
```
gilbertchen: The check here can be moved to outside the loop to avoid unnecessary loop under the dry run mode.

src/duplicacy_snapshotmanager.go
```diff
@@ -2227,6 +2231,9 @@ func (manager *SnapshotManager) DownloadFile(path string, derivationKey string)
 
 // UploadFile uploads a non-chunk file from the storage.
 func (manager *SnapshotManager) UploadFile(path string, derivationKey string, content []byte) bool {
+    if manager.config.dryRun {
```
gilbertchen: I think this check isn't necessary.  UploadFile is called by UploadSnapshot, and there is already a check there.

src/duplicacy_snapshotmanager.go
```diff
@@ -2215,9 +2217,11 @@ func (manager *SnapshotManager) DownloadFile(path string, derivationKey string)
         return nil
     }
 
-    err = manager.snapshotCache.UploadFile(0, path, manager.fileChunk.GetBytes())
-    if err != nil {
-        LOG_WARN("DOWNLOAD_FILE_CACHE", "Failed to add the file %s to the snapshot cache: %v", path, err)
+    if !manager.config.dryRun {
```
gilbertchen: It might be ok to save the downloaded file in the snapshot cache in dry-run mode.

src/duplicacy_backupmanager.go
```diff
@@ -1689,12 +1696,14 @@ func (manager *BackupManager) CopySnapshots(otherManager *BackupManager, snapsho
 
     for _, snapshot := range snapshots {
         if revisionMap[snapshot.ID][snapshot.Revision] == false {
-            continue
+          continue
+        }
+        if !manager.config.dryRun {
```
gilbertchen: CopySnapshots is used by the copy command only, so don't need to check the dry-run mode.

src/duplicacy_backupmanager.go
```diff
@@ -630,7 +634,9 @@ func (manager *BackupManager) Backup(top string, quickMode bool, threads int, ta
     }
     skippedFiles = append(skippedFiles, fileReader.SkippedFiles...)
 
-    manager.SnapshotManager.CleanSnapshotCache(localSnapshot, nil)
+    if !manager.config.dryRun {
```
gilbertchen: Glad you got this one.  CleanSnapshotCache is mainly to keep the last snapshot in the cache while deleting files from older snapshots, so we shouldn't really call it in dry-run mode.

src/duplicacy_backupmanager.go
```diff
@@ -1048,7 +1054,7 @@ func (manager *BackupManager) UploadSnapshot(chunkMaker *ChunkMaker, uploader *C
                 if _, found := chunkCache[chunkID]; found {
                     completionFunc(chunk, 0, true, chunk.GetLength(), 0)
                 } else {
-                    uploader.StartChunk(chunk, len(sequence))
+                     uploader.StartChunk(chunk, len(sequence))
```
tbain98: What's this doing here? It looks like a formatting-only change that makes it not line up with line 1055.

src/duplicacy_backupmanager.go
```diff
@@ -1689,7 +1696,7 @@ func (manager *BackupManager) CopySnapshots(otherManager *BackupManager, snapsho
 
     for _, snapshot := range snapshots {
         if revisionMap[snapshot.ID][snapshot.Revision] == false {
-            continue
+          continue
```
tbain98: What's this doing here? It looks like a formatting-only change that makes it not line up with the rest of the indentation levels.

src/duplicacy_chunkuploader.go
```diff
@@ -134,10 +134,12 @@ func (uploader *ChunkUploader) Upload(threadIndex int, task ChunkUploadTask) boo
...
-        return false
+    if !uploader.config.dryRun {
+        err = uploader.storage.UploadFile(threadIndex, chunkPath, chunk.GetBytes())
+        if err != nil {
+            LOG_ERROR("UPLOAD_CHUNK", "Failed to upload the chunk %s: %v", chunkID, err)
+            return false
+        }
     }
 
     LOG_DEBUG("CHUNK_UPLOAD", "Chunk %s has been uploaded", chunkID)
```
tbain98: Shouldn't this be inside the if? We really shouldn't log that we uploaded a chunk if we didn't actually do an upload.

gilbertchen:
Thank you for your contribution.  Sorry for taking so long to review and merge this PR.

PR #181, opened 2017-09-11
futher optimization for the copy command

jt70471:
This patch is a further optimization improvement for copy processing. The first optimization considered only chunks that existed in destination snapshot revisions and if the snapshot id were different, then the chunk would still attempt to be copied to the destination.

This patch is inherently different, as it considers all chunks at the destination, regardless of snapshot id or revision info. This is important because copy to cloud providers will take most users a very long time to complete. If the copy fails for any reason before the copy is completed and snapshot id / revisions are never written to the destination, a subsequent restart of the copy would try to re-copy chunks that had already been copied.

This optimization will ensure that chunks are skipped if found on the destination storage, regardless of whether or not they are contained in a snapshot id and revision.

### DIFF
--- a/integration_tests/copy_test.sh
+++ b/integration_tests/copy_test.sh
@@ -11,11 +11,21 @@ ${DUPLICACY} add -copy default secondary integration-tests $SECONDARY_STORAGE
 add_file file1
 add_file file2
 ${DUPLICACY} backup 
-${DUPLICACY}  copy -from default -to secondary
+${DUPLICACY} copy -from default -to secondary
 add_file file3
 add_file file4
 ${DUPLICACY} backup 
-${DUPLICACY}  copy -from default -to secondary
-${DUPLICACY} check --files -stats
+${DUPLICACY} copy -from default -to secondary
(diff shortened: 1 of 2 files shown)


gilbertchen:
Thanks.   ListAllFiles was exactly what I thought we could use as an alternative to checking existing snapshots one by one.  The only potential issue I can think of is that ListAllFiles may be slower on SFTP servers and networked drives (because of the hierarchical chunk directory), especially if there are only a limited number of snapshots.  This should be good for now but let us keep an eye on it.

jt70471:
I am using both an external USB and NAS share that use the hierarchical chunk directory structure, so I'll be able to assess how this performs as my snapshot revisions grow.

Took me a while to figure out that the chunkID is unique for each storage, even when added with the -copy to make them compatible... My first attempt to use the chunk file names wasn't finding any matches, then i figured out that I could use the chunkHash and GetChunkIDFromHash, all I had to do was ensure I used the otherManager for the otherChunkID... :)

Issue #182, opened 2017-09-11
Prune issue - chunk can't be found


jt70471:
Hi @gilbertchen, I believe I may have found a bug related to prune processing. At the very end of the integration-tests/copy_test.sh script, add the following:

${DUPLICACY} prune -id integration-tests -r 1-2 -exhaustive -exclusive -storage default

This should remove the two revisions made by the two backup commands. The attached output shows that indeed snapshot revisions 1 and 2 are deleted, but then after it would seem that prune should be finished, the oddball "can't be found" message is shown. Here's the snippet:

```
The snapshot integration-tests at revision 1 has been removed
The snapshot integration-tests at revision 2 has been removed
File /Users/jefft/DUPLICACY_TEST_ZONE/TEST_STORAGE/chunks/b3/1c/60e7eaca42d36488b50281549c42d0082034640b07fc975d4a5b91a5184c can't be found: lstat /Users/jefft/DUPLICACY_TEST_ZONE/TEST_STORAGE/chunks/b3/1c/60e7eaca42d36488b50281549c42d0082034640b07fc975d4a5b91a5184c: no such file or directory
File /Users/jefft/DUPLICACY_TEST_ZONE/TEST_STORAGE/chunks/b3/1c/60e7eaca42d36488b50281549c42d0082034640b07fc975d4a5b91a5184c.fsl can't be found: lstat /Users/jefft/DUPLICACY_TEST_ZONE/TEST_STORAGE/chunks/b3/1c/60e7eaca42d36488b50281549c42d0082034640b07fc975d4a5b91a5184c.fsl: no such file or directory
Chunk b31c60e7eaca42d36488b50281549c42d0082034640b07fc975d4a5b91a5184c can't be found
```
The chunk is in both snapshot revisions, so makes sense that once both revisions are removed, the chunk should have been deleted.

The only place I see this exact message being issued is in duplicacy_chunkdownloader.go, which makes no sense given that this was a prune command, not sure why this chunk would attempt to be downloaded, etc.

I've started looking into it, but I figured I'd report it as I'm sure you are much more familiar with why this may be occurring.

Any questions, please let me know. Thanks. The full output of the prune command is attached 
[here](https://github.com/gilbertchen/duplicacy/files/1293419/a.txt).

gilbertchen:
[...]

PR #183, opened 2017-09-11
fix prune bug when last snapshot is removed for issue #182

jt70471:
This prevents the bug when CleanSnapshotCache is called and the last snapshot pointer is not nil. Please review and let me know if you had a different approach in mind how to resolve this.

### DIFF
--- a/src/duplicacy_snapshotmanager.go
+++ b/src/duplicacy_snapshotmanager.go
@@ -2065,7 +2065,7 @@ func (manager *SnapshotManager) PruneSnapshots(selfID string, snapshotID string,
                          snapshot.ID, snapshot.Revision)
             }
             manager.snapshotCache.DeleteFile(0, snapshotPath)
-            fmt.Fprintf(logFile, "Deleted snapshot %s at revision %d\n", snapshot.ID, snapshot.Revision)
+            fmt.Fprintf(logFile, "Deleted cached snapshot %s at revision %d\n", snapshot.ID, snapshot.Revision)
         }
     }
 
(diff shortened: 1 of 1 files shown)


gilbertchen:
The change on latestSnapshot is correct but I'm concerned about allSnapshots being replaced by remainingSnapshots.  It shouldn't be touched because otherwise the local copies of snapshot files deleted by the prune command will remain in the cache.

jt70471:
My reasoning was that allSnapshots contains snapshots that have now been removed. That may or may not be an issue, but that's why I chose to replace the call to manager.ClearSnapshotCache with the remainingSnapshots instead of allSnapshots.

Are you saying that the patch looks good if only change the call to manager.CleanSnapshotCache to use allSnapshots instead of remainingSnapshots, i.e.:

```
manager.CleanSnapshotCache(latestSnapshot, allSnapshots)
```

Thanks

gilbertchen:
```
manager.CleanSnapshotCache(latestSnapshot, allSnapshots)
```
is right.  And then you don't even need to create `remainingSnapshots` at all -- just check `Flag` of `latestSnapshot` to see if it has been deleted, and set `latestSnapshot` to nil if so.

jt70471:
Ok, I think I finally got it... :)

One question though, I'm not clear on the CleanSnapshotCache call where latestSnapshot is using selfID. The snapshot id and revision being pruned could be for a different snapshot id than selfID, correct?

I'm probably just not understanding why selfID is used to determine the latest snapshot, so if you could shed some light I'd appreciate it. Thanks

gilbertchen:
selfID is the id of the current repository (from where the prune command is run).  Yes, the snapshot id being pruned could be different than selfID.  latestSnapshot needs to be the last snapshot of selfID, because we want to keep in the cache the chunks that made up the complete last snapshot (file list + chunk list, not file content), so that for the next backup, the last snapshot can be retrieved from the cache.  For all other snapshots only the chunks that can be used to construct the chunk list (to determine the set of referenced chunks for each snapshot) are kept in the cache.

jt70471:
Thanks for the explanation. I think the patch is ready for your review. Let me know if you have questions or would like another revision.

gilbertchen:
Perfect.  Thank you for your work!

Issue #175, opened 2017-09-10
Support regex in include/exclude patterns


tbain98:
Please support regexes in include/exclude patterns.

I know you've said you plan to do this eventually, so this issue is to 1) document that intent, and 2) request that you do it sooner rather than later.

jt70471:
@gilbertchen, did you have an idea how you'd like to implement this? I assume you'd want existing include/exclude processing to continue to work ASIS for compatibility reasons. Perhaps a new method of providing include/exclude filters using a new file named something like includeexclude or regularexpressions, etc?

That would give people a method of migrating from old to new include/exclude processing if they chose. Another approach would be to use the existing filters file and use a prefix like re:+ and re:- for example.

I'd be happy to tackle this one if you share how you'd like to see it implemented. Thanks.

gilbertchen:
re:+ and re:- should work.  Or maybe i: for include and e: for exclude.

We can even offer a -regex switch to the backup and restore commands to force all regex-based patterns.

I think you just need to add some regex matching code to [MatchPath](https://github.com/gilbertchen/duplicacy/blob/d1817ae5573499105f469a1d386c21012b9c528e/src/duplicacy_utils.go#L289).

jt70471:
This can be closed after inclusion of PR #187.

PR #187, opened 2017-09-12
add regex matching to include/exclude filters

jt70471:
Implement regular expression support for enhancement request #175. Can be used to replace existing filter functionality or in conjunction with existing functionality.

Regex include filters are preceded with "i:" and regex exclude filters preceded with "e:". Both backup (.duplicacy/filters) and restore command line arguments have been tested/verified.

Please review and consider for adoption to duplicacy. Thanks!

### DIFF
--- a/duplicacy/duplicacy_main.go
+++ b/duplicacy/duplicacy_main.go
@@ -679,6 +679,8 @@ func restoreRepository(context *cli.Context) {
     var patterns [] string
     for _, pattern := range context.Args() {
 
+        pattern = strings.TrimSpace(pattern)
+
         for strings.HasPrefix(pattern, "--") {
             pattern = pattern[1:]
         }
@@ -687,12 +689,19 @@ func restoreRepository(context *cli.Context) {
             pattern = pattern[1:]
         }
 
-        if pattern[0] != '+' && pattern[0] != '-' {
+        if duplicacy.IsUnspecifiedFilter(pattern) {
             pattern = "+" + pattern
         }
 
-        if pattern == "+" || pattern == "-" {
+        if duplicacy.IsEmptyFilter(pattern) {
             continue
+        }
+
+        if strings.HasPrefix(pattern, "i:") || strings.HasPrefix(pattern, "e:") {
+            valid, err := duplicacy.IsValidRegex(pattern[2:])
+            if  !valid || err != nil {
+                duplicacy.LOG_ERROR("SNAPSHOT_FILTER", "Invalid regular expression encountered for filter: \"%s\", error: %v", pattern, err)
+            }
         }
 
         patterns = append(patterns, pattern)
--- a/src/duplicacy_snapshot.go
+++ b/src/duplicacy_snapshot.go
@@ -67,7 +67,7 @@ func CreateSnapshotFromDirectory(id string, top string) (snapshot *Snapshot, ski
     }
 
     var patterns []string
-    
+
     patternFile, err := ioutil.ReadFile(path.Join(GetDuplicacyPreferencePath(), "filters"))
     if err == nil {
         for _, pattern := range strings.Split(string(patternFile), "\n") {
@@ -76,12 +76,23 @@ func CreateSnapshotFromDirectory(id string, top string) (snapshot *Snapshot, ski
                 continue
             }
 
-            if pattern[0] != '+' && pattern[0] != '-' {
+            if pattern[0] == '#' {
+                continue
+            }
+
+            if IsUnspecifiedFilter(pattern) {
                 pattern = "+" + pattern
             }
 
-            if pattern == "+" || pattern == "-" {
+            if IsEmptyFilter(pattern) {
                 continue
+            }
+
+            if strings.HasPrefix(pattern, "i:") || strings.HasPrefix(pattern, "e:") {
+                valid, err := IsValidRegex(pattern[2:])
+                if  !valid || err != nil {
+                    LOG_ERROR("SNAPSHOT_FILTER", "Invalid regular expression encountered for filter: \"%s\", error: %v", pattern, err)
+                }
             }
 
             patterns = append(patterns, pattern)
--- a/src/duplicacy_utils.go
+++ b/src/duplicacy_utils.go
@@ -39,6 +39,31 @@ func CreateRateLimitedReader(content []byte, rate int) (*RateLimitedReader) {
     }
 }
 
+func IsEmptyFilter(pattern string) bool {
+    if pattern == "+" || pattern == "-" || pattern == "i:" || pattern == "e:" {
+        return true
+    } else {
+        return false
+    }
+}
+
+func IsUnspecifiedFilter(pattern string) bool {
+    if pattern[0] != '+' && pattern[0] != '-' && pattern[0] != 'i' && pattern[0] != 'e' {
+        return true
+    } else {
+        return false
+    }
+}
+
+func IsValidRegex(pattern string) (valid bool, err error) {
+    _, err = regexp.Compile(pattern)
+    if err != nil {
+        return false, err
+    } else {
+        return true, err
+    }
+}
+
 func (reader *RateLimitedReader) Length() (int64) {
     return int64(len(reader.Content))
 }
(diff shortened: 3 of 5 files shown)


src/duplicacy_snapshot.go
```diff
@@ -76,19 +77,28 @@ func CreateSnapshotFromDirectory(id string, top string) (snapshot *Snapshot, ski
                 continue
             }
 
-            if pattern[0] != '+' && pattern[0] != '-' {
+            if pattern[0] == '#' {
+                continue
+            }
+
+            if pattern[0] != '+' && pattern[0] != '-' && pattern[0] != 'i' && pattern[0] != 'e' {
```
tbain98: From here down appears to duplicate the equivalent code in duplicacy_main.go. How about pulling out a helper method to avoid having to change things in multiple places in the future?

src/duplicacy_utils.go
```diff
@@ -300,6 +300,25 @@ func MatchPath(filePath string, patterns [] string) (included bool) {
...
+            if err != nil {
+                LOG_ERROR("SNAPSHOT_MATCH", "Error during regexp match: %s - %v", pattern, err)
+            }
+            if matched {
+                LOG_TRACE("SNAPSHOT_MATCH", "Regex include filePath=\"%s\", pattern=\"%s\", matched=%t", filePath, pattern[2:], matched)
+                return true
+            }
+        } else if strings.HasPrefix(pattern, "e:") {
+            allIncludes = false
+            matched, err := regexp.MatchString(pattern[2:], filePath)
```
tbain98: The code from here down is duplicate except for the return value, which would be easy enough to abstract out.

duplicacy/duplicacy_main.go
```diff
@@ -687,14 +689,21 @@ func restoreRepository(context *cli.Context) {
...
 
-        if pattern == "+" || pattern == "-" {
+        if pattern == "+" || pattern == "-" || pattern == "i:" || pattern == "e:" {
             continue
         }
 
+        if strings.HasPrefix(pattern, "i:") || strings.HasPrefix(pattern, "e:") {
+            _, err := regexp.Compile(pattern[2:])
+            if err != nil {
+                duplicacy.LOG_ERROR("SNAPSHOT_FILTER", "Invalid regular express encountered for filter: \"%s\", error: %v", pattern, err)
```
tbain98: Is logging the error sufficient here? Do we really want to just ignore the invalid regex and proceed as though everything else was the sum total of what we were asked to do?

I think we'd want to just stop the operation outright, to avoid doing something invalid/incorrect (which might cost the user money, and/or might give them a false sense that everything was working when it wasn't).

jt70471: In my testing, this does stop the operation after the invalid regex. Are you finding that not to be the case?

src/duplicacy_utils.go
jt70471: This is consistent with the existing pattern matching code for the + and - filter prefixes, so I don't feel that is needs to be reworked.

jt70471:
The PR has been updated with the requested changes, except for the remaining comment listed above.

src/duplicacy_snapshot.go
jt70471: done

src/duplicacy_snapshot.go
```diff
@@ -76,14 +76,24 @@ func CreateSnapshotFromDirectory(id string, top string) (snapshot *Snapshot, ski
...
+            if IsUnspecifiedFilter(pattern) {
                 pattern = "+" + pattern
             }
 
-            if pattern == "+" || pattern == "-" {
+            if IsEmptyFilter(pattern) {
                 continue
             }
 
+            valid, err := IsValidRegex(pattern)
```
tbain98: Doesn't this line demand that all patterns be regexes? That's not what we want.

duplicacy/duplicacy_main.go
```diff
@@ -687,14 +689,20 @@ func restoreRepository(context *cli.Context) {
...
+        if duplicacy.IsUnspecifiedFilter(pattern) {
             pattern = "+" + pattern
         }
 
-        if pattern == "+" || pattern == "-" {
+        if duplicacy.IsEmptyFilter(pattern) {
             continue
         }
 
+        valid, err := duplicacy.IsValidRegex(pattern)
```
tbain98: Same question here about how this code appears to be requiring all patterns to be regexes.

duplicacy/duplicacy_main.go
tbain98: My comments were based purely on reading the code not running it, and I'm still coming up to speed on Go, so I very well could have misinterpreted something. I would have expected there to be a return after the LOG_ERROR call, which was what I was commenting based on, but I haven't attempted to actually test the code by running it.

src/duplicacy_snapshot.go
```diff
@@ -76,14 +76,24 @@ func CreateSnapshotFromDirectory(id string, top string) (snapshot *Snapshot, ski
...
+
+            if  !valid || err != nil {
+                LOG_ERROR("SNAPSHOT_FILTER", "Invalid regular express encountered for filter: \"%s\", error: %v", pattern, err)
```
tbain98: I've still got the same concern as before: I don't see how logging a problem with a single regex results in preventing all the rest from being (wrongly) applied. A set of unit tests (https://golang.org/pkg/testing/) would go a long way towards proving correct operation in this case.

duplicacy/duplicacy_main.go
jt70471: I don't know go either, but I definitely confirmed that LOG_ERROR will stop the program after logging the message.

jt70471:
OK, the filter is now checked for the "i:" or "e:" prefix for regular expression filters before trying to validate the regex.

tbain98:
The changes to only test for regex validity when it's an actual regex address the remaining concern I had. Thanks for the fixes.

BTW, your explanation of exiting prompted me to look at the code for LOG_ERROR, and I discovered that @gilbertchen has implemented code that panics if something is logged at a level > WARN. That's not a paradigm I've come across before (especially because usually only FATAL is considered fatal, not ERROR), but at least now I'm aware of it for the future. Thanks.

src/duplicacy_snapshot.go
jt70471: oops, yes, has been fixed in latest PR.

duplicacy/duplicacy_main.go
jt70471: oops, yes, has been fixed in latest PR.

src/duplicacy_snapshot.go
jt70471: I believe you are now comfortable that LOG_ERROR terminates the program after logging the error. It's used throughout the duplicacy code, so I believe it's usage here is appropriate.

tbain98: That's correct, I have no remaining concerns about the changes in this pull request.

TheBestPessimist:
Is it possible to also add an explanation in the GUIDE.md file?
This way everyone would know that there are 2 different regex pattern styles, which do different things (eventually mark the old one as deprecated, and the new one (this one) as the correct one)

tbain98:
GUIDE.md is one of the five files Jeff modified under this PR. Was there
specific content you wanted to see that he didn't cover?

On Sep 14, 2017 1:36 AM, "Cristian Viorel Pasat" <notifications@github.com>
wrote:

Is it possible to also add an explanation in the GUIDE.md file?
This way everyone would know that there are 2 different regex pattern
styles, which do different things (eventually mark the old one as
deprecated, and the new one (this one) as the correct one)

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/gilbertchen/duplicacy/pull/187#issuecomment-329399619>,
or mute the thread

jt70471:
I did update GUIDE.md as @tbain98 stated. As far as regular expressions ultimately replacing the existing wildcard matching, that would be up to the project maintainers.

While I will use RE's exclusively, they're not for everyone as the syntax can be quite challenging (and powerful), so my guess is that the original pattern matching should stick around, but not my call.

gilbertchen:
Originally LOG_ERROR was meant to report errors that may be recoverable (if a customized logging function is provided), so a return statement is preferred after LOG_ERROR, but I failed to adhere to this rule throughout the development.  It is now safe to assume that LOG_ERROR should be made irrecoverable and that the statement after LOG_ERROR will not be executed.

My only concern with this PR is that the regular expressions could have been precompiled for better performance.  However, that would make the calling flow more complicated and I prefer code simplicity unless people start to complain about performance.

jt70471:
I did consider saving the pre-compiled RE's rather, however, my lack of go experience got in the way. If you could give some hints how/where to store a slice of regexp pointers, I'd be happy to implement that optimization. Thanks.

TheBestPessimist:
Ah, please excuse me, i don't know how to use github apparently. If i look on the top at "files changed" some files were hidden because (unknown) reasons. Now i can see the change in the guide file. Please ignore my previous comment :blush:

hirowatari:
When will this be released?  Just wondering because I thought it already was and was trying to debug why my rules weren't working.

jt70471:
Hi, the patch has been merged into the source, but this was done after the 2.0.9 binary package was built. You'll have to wait until @gilbertchen releases a new binary package or build the binary yourself from the source.

PR #191, opened 2017-09-15
store compiled regex patterns for performance optimization

jt70471:
Hi, here's my attempt at optimizing the regex filter performance. It works as follows:

1. When a filter is first compiled to verify its validity, the Regexp pointer returned by the compile is stored in a map for future reference.

2. MatchPath is updated to look for the presence of a compiled regex in the map. It uses it if stored and will compile/use the regex pattern if not found in the map, although in testing, I did not encounter a scenario where the already compiled regex was not available in the map.

3. I had no idea where/how to store the compiled regex map, so I created a new src/duplicacy_regex.go that simply has the variable declaration for the map. I suspect this is the go equivalent of a global variable, which I am not in the habit of doing, but I was not able to figure out where to store the map that would be accessible since the MatchPath and IsValidRegex are utlity routines that are not associated with an object ptr, etc.

4. Before storing a new compiled regex to the map, I verify if the map already contains the compiled regex and LOG_WARN if the entry already exists. This should perhaps be a LOG_ERROR, as the implication is that the user has duplicate include or exclude filters (minus the prefix). I decided on the warning because I felt that even if the user has duplicate filters, matching will stop on the first match whether it's an include or exclude filter, so the fact that there's another filter statement later in the filters file, I would not change the match/nomatch. 

5. I left in multiple LOG_DEBUG calls, which I would probably remove if you are OK with the patch, just let me know and I'll remove those debug statements.

6. I tested this for both backup/restore and in my testing, the compiled regex statements are used successfully, avoiding having to compile the regex for each comparison, which can of course be frequent for large filesystems.

If there's a better way to accomplish this, any input you can provide I'd be happy to implement the way you want.

### DIFF
--- a/src/duplicacy_utils.go
+++ b/src/duplicacy_utils.go
@@ -31,6 +31,16 @@ type RateLimitedReader struct {
     StartTime time.Time
 }
 
+var RegexMap map[string]*regexp.Regexp
+
+func init() {
+
+    if RegexMap == nil {
+        RegexMap = make(map[string]*regexp.Regexp)
+    }
+
+}
+
 func CreateRateLimitedReader(content []byte, rate int) (*RateLimitedReader) {
     return &RateLimitedReader {
         Content: content,
@@ -56,10 +66,20 @@ func IsUnspecifiedFilter(pattern string) bool {
 }
 
 func IsValidRegex(pattern string) (valid bool, err error) {
-    _, err = regexp.Compile(pattern)
+
+    var re *regexp.Regexp = nil
+
+    if re, valid = RegexMap[pattern]; valid && re != nil {
+        return true, nil
+    }
+
+    re, err = regexp.Compile(pattern)
+
     if err != nil {
         return false, err
     } else {
+        RegexMap[pattern] = re
+        LOG_DEBUG("REGEX_STORED", "Saved compiled regex for pattern \"%s\", regex=%#v", pattern, re)
         return true, err
     }
 }
@@ -313,9 +333,13 @@ func matchPattern(text string, pattern string) bool {
 // include patterns, and included otherwise.
 func MatchPath(filePath string, patterns [] string) (included bool) {
 
-    allIncludes := true
-    for _, pattern := range patterns {
+    var re *regexp.Regexp = nil
+    var found bool
+    var matched bool
 
+    allIncludes := true
+
+    for _, pattern := range patterns {
         if pattern[0] == '+' {
             if matchPattern(filePath, pattern[1:]) {
                 return true
@@ -325,24 +349,23 @@ func MatchPath(filePath string, patterns [] string) (included bool) {
              if matchPattern(filePath, pattern[1:]) {
                 return false
             }
-        } else if strings.HasPrefix(pattern, "i:") {
-            matched, err := regexp.MatchString(pattern[2:], filePath)
-            if err != nil {
-                LOG_ERROR("SNAPSHOT_MATCH", "Error during regexp match: %s - %v", pattern, err)
+        } else if strings.HasPrefix(pattern, "i:") || strings.HasPrefix(pattern, "e:") {
+            if re, found = RegexMap[pattern[2:]]; found {
+                matched = re.MatchString(filePath)
+            } else {
+                re, err := regexp.Compile(pattern)
+                if err != nil {
+                    LOG_ERROR("REGEX_ERROR", "Invalid regex encountered for pattern \"%s\" - %v", pattern[2:], err)
+                }
+                RegexMap[pattern] = re
+                matched = re.MatchString(filePath)
             }
             if matched {
-                LOG_TRACE("SNAPSHOT_MATCH", "Regex include comparison for filePath=\"%s\", pattern=\"%s\", matched=%t", filePath, pattern[2:], matched)
-                return true
-            }
-        } else if strings.HasPrefix(pattern, "e:") {
-            allIncludes = false
-            matched, err := regexp.MatchString(pattern[2:], filePath)
-            if err != nil {
-                LOG_ERROR("SNAPSHOT_MATCH", "Error during regexp match: %s - %v", pattern, err)
-            }
-            if matched {
-                LOG_TRACE("SNAPSHOT_MATCH", "Regex exclude comparison for filePath=\"%s\", pattern=\"%s\", matched=%t", filePath, pattern[2:], matched)
-                return false
+                return strings.HasPrefix(pattern, "i:")
+            } else {
+                if strings.HasPrefix(pattern, "e:") {
+                    allIncludes = false
+                }
             }
         }
     }
--- a/duplicacy/duplicacy_main.go
+++ b/duplicacy/duplicacy_main.go
@@ -708,6 +708,8 @@ func restoreRepository(context *cli.Context) {
 
     }
 
+    duplicacy.LOG_DEBUG("REGEX_DEBUG", "There are %d compiled regular expressions stored", len(duplicacy.RegexMap))
+
     storage.SetRateLimits(context.Int("limit-rate"), 0)
     backupManager := duplicacy.CreateBackupManager(preference.SnapshotID, storage, repository, password)
     duplicacy.SavePassword(*preference, "password", password)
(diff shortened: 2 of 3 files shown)


jt70471:
On second thought, I think I could eliminate duplicacy_regex.go altogether and just create the map in duplicacy_utils.go since that's the only place it's referenced and add an init() function to utils to initialize the map on first use.

I'll wait for @gilbertchen to comment before making any updates though as I may have completed missed the mark. :)

gilbertchen:
I think we should avoid storing the precompiled regexes in a map -- a lookup in the map may be more costly than constructing a regex.  Instead, we should pass a slice of precompiled regexes along with the patterns.  For example, the MatchPath function should become:

```
func MatchPath(filePath string, patterns [] string, regexes [] *regexp.Regexp) (included bool) {
```
`regexes` should have the same size as `patterns`, and initially contains nil pointers.  For each regex pattern, `MatchPath` will check if the corresponding element of `regexes` is nil or not.  If it is nil, compile the regex otherwise just use it.  This guarantees constant lookup times.

It will be the caller who provides `regexes`.  However, I didn't check the call graph thoroughly to see if this is feasible in all cases.

tbain98:
@gilbertchen Lookups in a hash-based map are O(1) as long as you have a reasonable hashing algorithm, and are orders of magnitude less costly than compiling a regex. Does your comment imply that Go doesn't provide a hash-based map?

tbain98:
I should have Googled before I responded.

https://blog.golang.org/go-maps-in-action indicates that the map implementation used by @jt70471's code is indeed a hash map. So I'm now fully in favor of leaving the method signature unchanged and storing the map within duplicacy_utils.go as Jeff proposed.

src/duplicacy_utils.go
```diff
@@ -56,10 +56,21 @@ func IsUnspecifiedFilter(pattern string) bool {
 }
 
 func IsValidRegex(pattern string) (valid bool, err error) {
-    _, err = regexp.Compile(pattern)
+
+    var re *regexp.Regexp = nil
+
+    if re, valid = RegularExpressions[pattern]; valid && re != nil {
+        LOG_WARN("REGEX_FOUND", "Found already compiled regex for pattern \"%s\", regexp=%#v", pattern, re)
```
tbain98: I understand the logic behind why you added this line, but it makes it such that this method can only be meaningfully used in the context of a single pass through a single group of regexes that are supposed to have only one instance of each one.

A better approach would be to test the raw strings for duplicates, leaving duplicate-detection in the hands of the code that has the full list and keeping this method focused solely on determining if the regex is valid (and populating the cache if it's a new regex).

The duplicate-detection in the calling method could be as simple as putting all values into a set and checking whether it has the same size as the original collection; fewer entries in the set means that there is at least one set of duplicates in the batch.

src/duplicacy_utils.go
```diff
@@ -325,24 +340,25 @@ func MatchPath(filePath string, patterns [] string) (included bool) {
...
-            matched, err := regexp.MatchString(pattern[2:], filePath)
-            if err != nil {
-                LOG_ERROR("SNAPSHOT_MATCH", "Error during regexp match: %s - %v", pattern, err)
+        } else if strings.HasPrefix(pattern, "i:") || strings.HasPrefix(pattern, "e:") {
+            if re, found = RegularExpressions[pattern[2:]]; found {
+                LOG_DEBUG("REGEX_FOUND", "Compiled regex for pattern \"%s\" found", pattern)
+                matched = re.MatchString(filePath)
+            } else {
+                LOG_DEBUG("REGEX_NOTFOUND", "Compiled regex for pattern \"%s\" not found!", pattern)
+                matched, _ = regexp.MatchString(pattern[2:], filePath)
```
tbain98: Why wouldn't we simply compile and store the regex, then use it? We should be caching regexes in all cases, not just some.

src/duplicacy_utils.go
```diff
@@ -325,24 +340,25 @@ func MatchPath(filePath string, patterns [] string) (included bool) {
...
             }
+            LOG_DEBUG("REGEX_RESULT", "Regex comparison for filePath=\"%s\", pattern=\"%s\", matched=%t", filePath, pattern, matched)
             if matched {
-                LOG_TRACE("SNAPSHOT_MATCH", "Regex exclude comparison for filePath=\"%s\", pattern=\"%s\", matched=%t", filePath, pattern[2:], matched)
-                return false
+                if strings.HasPrefix(pattern, "i:") {
```
tbain98: Just `return strings.HasPrefix(pattern, "i:")` instead of the complicated if statement.

src/duplicacy_utils.go
jt70471: The message is simply meant to convey that we've already encountered this same regex pattern before, which may (probably) indicates that the user has duplicate filters specified. I made it a warning instead of error, because after the first one is matched, no further filters are matched, so it shouldn't affect the overall logic, etc.

Mostly it was just a sanity check I put in to verify what I was doing with the compiled regex's and not re-compiling them over and over, etc.

src/duplicacy_utils.go
jt70471: I see your point, but again, I was thinking defensively how to ensure that I'd handle different scenarios. By the time MatchPath is called, all of the filters should have been compiled and stored, by calling IsValidRegex and that is indeed the case as of this PR.

In future updates, if code is introduced that doesn't call IsValidRegex first, then the code here would handle either these where we have already compiled a regex or if for some as of now unknown reason we don't have a compiled regex, it's compiled and stored, etc.

src/duplicacy_utils.go
jt70471: duh, thanks... :)

src/duplicacy_utils.go
tbain98: While you were writing the code, I can see using WARN as the logging level. But now that you've got it working, it should drop to DEBUG.

The method simply determines if the regex is valid, caching it if it's not already cached. Nothing about that logic says that seeing the same regex more than once (over the entire lifetime of the program) is invalid or worthy of a WARN/ERROR. Only the code that's assembling the list of regexes to be evaluated cares whether you've seen this regex before, so if anyone's going to be throwing WARNs/ERRORs, it's that method.

So if you care about checking for (and logging) that condition, add it to that method. If not, then don't worry about it. But either way, make this line a DEBUG, because this method shouldn't care about the business logic of that other method.

src/duplicacy_utils.go
tbain98: You're right that handling the case where the regex isn't in the cache is defensive programming, but what I'm recommending is defensive programming at the next level up: ensuring that you not only do something, but do the right something, if you hit that condition.

You're baking logic for your calling algorithms into these low-level methods, which couples them more tightly than they should be and hinders the ability to reuse them if the need arose. Why should you treat this particular regex any different than any others, just because someone happens not to have called IsValidRegex() first (or just because all the code that's been written so far calls IsValidRegex() first)? And why should you bake in inconsistent behavior now just because you know it doesn't currently matter either way, when it takes almost no more effort to do it the right way and then know that future maintainers won't have to worry about it if they implement something that doesn't call IsValidRegex() first?

jt70471: Changed my mind and have reworked this to compile, store, and used the compiled regex if not already found in the map.

tbain98: Sounds great. It appears our comments were ships passing in the night, but I'm glad we both ended up at the same place.

src/duplicacy_utils.go
jt70471: I think I'd prefer to just remove the message altogether.

tbain98: I'm good with that.

jt70471: it's gone, i think that addresses all of your comments.

tbain98: The only remaining comment I had was that I agreed with your suggestion to move the map from duplicacy_regex.go to duplicacy_utils.go and get rid of duplicacy_regex.go.

jt70471:
I have moved the regex map into duplicacy_utils.go and removed duplicacy_regex.go.

tbain98:
I'm good with this PR. Thanks for making a change that will have a positive
performance impact compared with the original regex implementation.

gilbertchen:
I'll accept this PR but I still have reservations about the use of the map -- in theory it is O(1) but you can never know its performance for sure in real worlds.  We can revisit this in the future if there is a performance issue.

jt70471:
Thanks, a slice will certainly perform slightly better, but the map will perform much better than having to compile the regex for each call to MatchPath.

Issue #179, opened 2017-09-11
B2 efficiency


fracai:
Let me preface this by saying I hope I got all my math right. Corrections are welcome. I'm attaching the python script I used for calculations and a table of those results.

tl;dr: **b2_list_file_names** is inefficient to check for existing chunks. Instead use **b2_download_file_by_name** with a header "Range: bytes=0-0".

Duplicacy calls **b2_list_file_names** prior to every upload to check if a given chunk has already been uploaded. As referenced in #30, the B2 pricing means around 10 GB can be uploaded for free each day, and then $.001/GB. However, the list query only returns up to 1000 entries each call so multiple calls per chunk must be made once more than 1000 chunks have been uploaded. 1000 chunks works out to around only 4 GB. By the time I reach 1 TB of data (around 263000 chunks) it's going to be 263 calls to **b2_list_file_names** for each new chunk. Ignoring the free data and transactions (negligible given the total data) it's going to cost almost $140 to make the 35 million calls while uploading. 10 TB would require 2.5 million chunks, 3.5 billion total calls, cost $13,750, and require another 2622 list calls just to upload another new chunk. That's over a penny every 4 MB.

Now, there is the **b2_get_file_info** call, but that requires passing the **fileId** which is normally obtained from **b2_list_file_names**. Duplicacy could save the **fileId** in the snapshots files, but that would require an architecture change.

Alternatively, Duplicacy could call **b2_download_file_by_name** and look at the response code (200 vs 400) and headers. A GET or HEAD adds the entire payload size to your download quota, but including a Range header only adds the requested Range size. Setting a "Range: bytes=0-0" is only a single byte transferred. This reduces the cost by an order of magnitude and also is guaranteed to require a single call, rather than multiple calls. This reduces the cost to upload 10 TB from $13,750 to $1.05. The additional single byte (GET or HEAD) equals around 2 billion calls for a penny. Conveniently, the response headers include the relevant info also returned by **b2_get_file_info**. It looks like the only real difference is a lack of the bucket id.

On the more hacky side, **b2_hide_file** and **b2_delete_file_version** are each unbilled calls. These could be combined to test for the existence of a file. I can see an issue if the process dies before unhiding the chunk and it looks like Duplicacy uses B2 hidden files for fossils. I think this option is more trouble than it's worth.

[...]

PR #209, opened 2017-09-22
B2 efficiency

fracai:
As discussed in #179 these commits use _b2_download_file_by_name_ instead of _b2_list_file_names_ to retrieve chunk information.
The switch only occurs when a single file is being checked without previous versions (fossils). The call to download doesn't return the "**B2 action**", so the code just supplies "**upload**". This matches the current expectation as a hidden chunk would not be visible to _b2_list_file_names_ either. A fossil would be visible to _b2_list_file_versions_, but that is only called if _includeVersions_ is **true**.
I haven't seen any noticeable impact to run time, but I can see the change in recorded transaction classes on the B2 summary.

The only part that feels outright hacky to me is the switch statement at the top of _call_. I made the http call a HEAD request (B2 could at some point might optimize their API to not charge the requested bytes) and thus needed to pass in a different _input_ paramater. I picked _bool_. This does work, but I wonder if it would be better supported by adding another paramater to specify the request type? Or an enum to make that switch statement more explicit?

### DIFF
--- a/src/duplicacy_b2client.go
+++ b/src/duplicacy_b2client.go
@@ -103,7 +103,7 @@ func (client *B2Client) call(url string, requestHeaders map[string]string, input
         case int:
             method = "GET"
             inputReader = bytes.NewReader([]byte(""))
-        case nil:
+        case bool:
             method = "HEAD"
             inputReader = bytes.NewReader([]byte(""))
         }
@@ -295,20 +295,28 @@ func (client *B2Client) ListFileNames(startFileName string, singleFile bool, inc
     input["bucketId"] = client.BucketID
     input["startFileName"] = startFileName
     input["maxFileCount"] = maxFileCount
+    inputMapBool := true
 
-    url := client.APIURL + "/b2api/v1/b2_list_file_names"
-    requestHeaders := map[string]string{}
-    if includeVersions {
-        url = client.APIURL + "/b2api/v1/b2_list_file_versions"
-    } else if singleFile {
-        // handle a single file with no versions as a special case to download the first byte of the file
-        url = client.APIURL + "/b2api/v1/b2_download_file_by_name"
-        requestHeaders["Range"] = "bytes=0-0"
-        // HEAD request
-        input = nil
-    }
     for {
-        readCloser, responseHeader, _, err := client.call(url, requestHeaders, input)
+        url := client.APIURL + "/b2api/v1/b2_list_file_names"
+        requestHeaders := map[string]string{}
+        if includeVersions {
+            url = client.APIURL + "/b2api/v1/b2_list_file_versions"
+        } else if singleFile {
+            // handle a single file with no versions as a special case to download the first byte of the file
+            url = client.DownloadURL + "/file/" + client.BucketName + "/" + startFileName
+            requestHeaders["Range"] = "bytes=0-0"
+            // HEAD request
+            inputMapBool = false
+        }
+        var readCloser io.ReadCloser
+        var responseHeader http.Header
+        var err error
+        if inputMapBool {
+            readCloser, responseHeader, _, err = client.call(url, requestHeaders, input)
+        } else {
+            readCloser, responseHeader, _, err = client.call(url, requestHeaders, false)
+        }
         if err != nil {
             return nil, err
         }
@@ -318,12 +326,13 @@ func (client *B2Client) ListFileNames(startFileName string, singleFile bool, inc
         output := B2ListFileNamesOutput {
         }
 
-        if singleFile && !includeVersions {
+        if !inputMapBool && singleFile && !includeVersions {
             // construct the B2Entry from the response headers of the download request
-            fileID := responseHeader.Get("X-Bz-File-Id")
-            fileName := responseHeader.Get("X-Bz-File-Name")
-            fileAction := responseHeader.Get("X-Bz-Action")
-            fileSize, _ := strconv.ParseInt(responseHeader.Get("Content-Length"), 0, 64)
+            fileID := responseHeader.Get("x-bz-file-id")
+            fileName := responseHeader.Get("x-bz-file-name")
+            fileAction := "upload"
+            rangeString := responseHeader.Get("Content-Range")
+            fileSize, _ := strconv.ParseInt(rangeString[strings.Index(rangeString, "/")+1:], 0, 64)
             fileUploadTimestamp, _ := strconv.ParseInt(responseHeader.Get("X-Bz-Upload-Timestamp"), 0, 64)
 
             return []*B2Entry {&B2Entry { fileID, fileName, fileAction, fileSize, fileUploadTimestamp }}, nil
--- a/src/duplicacy_b2client.go
+++ b/src/duplicacy_b2client.go
@@ -84,14 +84,14 @@ func (client *B2Client) retry(backoff int, response *http.Response) int {
 func (client *B2Client) call(url string, method string, requestHeaders map[string]string, input interface{}) (io.ReadCloser, http.Header, int64, error) {
 
 	switch method {
-		case http.MethodGet:
-        	break
-        case http.MethodHead:
-        	break
-        case http.MethodPost:
-        	break
-        default:
-        	return nil, nil, 0, fmt.Errorf("unhandled http request method: "+method)
+	case http.MethodGet:
+		break
+	case http.MethodHead:
+		break
+	case http.MethodPost:
+		break
+	default:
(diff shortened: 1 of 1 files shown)


gilbertchen:
Thank you for the work.  I will review the changes tomorrow and may take a couple of days to run some tests.

fracai:
Sure, let me know if you have any architecture preferences. After putting this together I had the thought that it might be better served in its own method, but figured that was getting more towards style preferences.
I hope it proves useful!

src/duplicacy_b2client.go
```diff
@@ -81,7 +81,7 @@ func (client *B2Client) retry(backoff int, response *http.Response) int {
 	return backoff
 }
 
-func (client *B2Client) call(url string, input interface{}) (io.ReadCloser, int64, error) {
+func (client *B2Client) call(url string, requestHeaders map[string]string, input interface{}) (io.ReadCloser, http.Header, int64, error) {
```
gilbertchen: It may be better to add the method as an argument and not pass requestHeaders.

src/duplicacy_b2client.go
```diff
@@ -115,6 +118,12 @@ func (client *B2Client) call(url string, input interface{}) (io.ReadCloser, int6
 			request.Header.Set("Authorization", client.AuthorizationToken)
 		}
 
+		if requestHeaders != nil {
```
gilbertchen: The header `Range: bytes=0-0` can be added here if the method is "GET".

src/duplicacy_b2client.go
```diff
@@ -131,11 +140,11 @@ func (client *B2Client) call(url string, input interface{}) (io.ReadCloser, int6
...
 				continue
 			}
-			return nil, 0, err
+			return nil, nil, 0, err
 		}
 
 		if response.StatusCode < 300 {
-			return response.Body, response.ContentLength, nil
+			return response.Body, response.Header, response.ContentLength, nil
 		}
```
gilbertchen: I think you need to check if the response code is 404 and the method is HEAD.  If so return nil httpHeader.  Without this check it will keep retrying when looking up a non-existent file.

src/duplicacy_b2client.go
```diff
@@ -299,6 +323,18 @@ func (client *B2Client) ListFileNames(startFileName string, singleFile bool, inc
 
 		output := B2ListFileNamesOutput{}
 
+		if !inputMapBool && singleFile && !includeVersions {
```
gilbertchen: Again here you need to check if responseHeader is nil.

src/duplicacy_b2client.go
fracai: Is there any desire to keep requestHeaders as an argument in the event that other calls need to pass anything in the future? Looking at the B2 API it looks like only b2_upload_file and b2_download_file_by_name take a request header other than authorization.

I'll make the method an enum and construct the header in that switch.

gilbertchen:
Here is how I run the tests.  First create a file named `test_storage.conf` under `.\src`:
```
{
    "b2": {
        "bucket": "bucket_name",
        "account": "b2_account",
        "key": "b2_key"
    }
}
```
Then run the following tests:

```
go test -run TestB2Client
go test -run TestStorage ./src -storage b2
go test -run TestBackupManager ./src -storage b2
```

src/duplicacy_b2client.go
```diff
@@ -284,13 +293,28 @@ func (client *B2Client) ListFileNames(startFileName string, singleFile bool, inc
 	input["bucketId"] = client.BucketID
 	input["startFileName"] = startFileName
 	input["maxFileCount"] = maxFileCount
+	inputMapBool := true
```
tbain98: I think this will disappear if you take my suggestion to pass the HTTP verb in as an argument, but this isn't a very approachable variable name for anyone seeing this code without a pre-existing understanding of what it does (not least because a boolean's name should represent what will happen if it is true, but this variable's name implies that you'll pass a boolean in place of the input map argument, but then does the opposite). I hope you'll be eliminating it entirely (setting method = "HEAD" and input = nil, for example), but if you were going to keep it, a better name would be something like passBoolToInputParameter.

src/duplicacy_b2client.go
tbain98: Do we need to test !inputMapBool here? Is there a situation where we'd get here and singleFile would be true, includeVersions would be false, and inputMapBool would be true? Seems like unnecessary complexity to the if statement.

src/duplicacy_b2client.go
```diff
@@ -94,19 +94,22 @@ func (client *B2Client) call(url string, input interface{}) (io.ReadCloser, int6
...
-				return nil, 0, err
+				return nil, nil, 0, err
 			}
 			inputReader = bytes.NewReader(jsonInput)
 		case []byte:
 			inputReader = bytes.NewReader(input.([]byte))
 		case int:
 			method = "GET"
 			inputReader = bytes.NewReader([]byte(""))
+		case bool:
```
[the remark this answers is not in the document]
fracai: You're right, there are the constants in the http package. Those should be used here. https://golang.org/pkg/net/http/#pkg-constants

More generally, I think this can be cleaner by passing in a map of headers, input as a bytes.Reader, and the method as string (a switch can restrict this to only match the http constants)

```
call(url string, requestMethod string, requestHeaders map[string]string, input *bytes.Reader)

…

switch requestMethod {
    default:
        return nil, nil, 0, err
    case MethodGet:
        break
    case MethodHead:
        break
    case MethodPost:
        break
}
```

src/duplicacy_b2client.go
fracai: all valid

src/duplicacy_b2client.go
fracai: You're right, it's there as a sanity check, but it is redundant. That bool is going away regardless. I'll see if it makes sense to have anything similar after the other changes.

fracai: I'll add that and some other error checking to be sure the expected headers are present.

src/duplicacy_b2client.go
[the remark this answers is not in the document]
tbain98: What you've proposed for the request method sounds fine given that strongly-typed enums don't exist in Go.

I'm not wild about the suggestion that creating the bytes.Reader should be done each place that the method is called; that feels like it'll lead to code duplication. Rather, keep the current parameter type of interface, update the logic to allow passing nil for the cases where empty content should be used for the request, and remove the special cases for int and bool (passing nil instead).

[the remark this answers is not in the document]
fracai: I started thinking about the duplication as well. One idea was making use of lambdas. I haven't had a chance to actually implement anything yet and lambdas seem heavy handed here as well. I'll try your suggestion.

fracai:
I've cleaned up the method specification and added a bunch of error checking.
I'm still passing in the requestHeaders as it really doesn't seem clean to construct those in the _call_ method, but I can if that's desired.
I haven't had a chance to run the formal tests yet, but it does seem fine based on a few manual runs.

src/duplicacy_b2client.go
```diff
@@ -144,15 +159,24 @@ func (client *B2Client) call(url string, input interface{}) (io.ReadCloser, int6
...
 			continue
 		} else if response.StatusCode == 403 {
 			if !client.TestMode {
-				return nil, 0, fmt.Errorf("B2 cap exceeded")
+				return nil, nil, 0, fmt.Errorf("B2 cap exceeded")
 			}
 			continue
+		} else if response.StatusCode == 404 {
+			if http.MethodHead == method {
+				return nil, nil, 0, fmt.Errorf("URL request '%s' returned status code %d", url, response.StatusCode)
```
gilbertchen: Can't return an error here.  Should instead tell the caller that the given file doesn't exist in the storage.

fracai: I'm not sure I follow. If I look at how the code is used, returning an error does indicate that the file doesn't exist.

b2storage.GetFileInfo for example

Are you saying it should just be swallowing the error and returning "nil, nil, 0, nil"?

gilbertchen: Yes, if the file to be looked up doesn't exist, GetFileInfo expects ListFileNames to return an empty Entry array, so maybe here should return a responseHeader of nil and in ListFileNames use that as the condition to return the empty Entry array.

fracai: I've updated the 404 branch to return a nil error and then ListFileNames treats that nil responseHeader as a missing file and returns an empty array.

gilbertchen:

Issue #163, opened 2017-09-06
Shipping local backups


nikjft:
Backing up is fairly resource intensive and supporting multiple destinations for each repo is also kind of tricky (and seemingly impossible with the GUI tool).

What I'd like to do is have Duplicacy back up to a network drive on my NAS, and then just copy those backups to a remote server.

If I do this, will I be able to restore directly from that remote server, or would I have to download all of the backup files in order to be able to restore? 

What if I did the same with another local drive (i.e. mirror my Duplicacy backups and rotate drives off-site)?

jt70471:
You should be able to do that I would think, you'd have to adjust your preferences file to point to the new "remote" location, but perhaps an easier way would be to "duplicacy add" the new remote destination, then you could "duplicacy copy" the snapshot id's and revision(s) you'd like to copy off-site.

I backup everything to a local (nas) share, then "duplicacy copy" from local to Backblaze b2. I also use "duplicacy prune" with different retention criteria for the local and b2 storage locations, so you really can tailor this to your needs. i.e. you can copy everything local to your cloud storage and use the same prune retention criteria if that's what you want or they can differ, it's totally up to you.

What I would avoid, however, is doing a "duplicacy backup" to both local and cloud storage, because each will have its own independent revision numbers. With the copy approach, the snapshot id's and revision numbers are consistent and copy runs faster than backup as an added benefit, and it's an exact copy of the backup, etc.

I have recently submitted an optimization patch for copy performance, where it copies only the new chunks to the destination storage. If/when this is included in duplicacy, it really speeds up the copy process after the initial copy is performed.

Good luck.

gilbertchen:
@jt70471 is right.  Adding a second storage and then running the copy command is the recommended way.

If you want to use another app (such as rclone) to copy local backups residing on a network drive to a remote server, then the remote server needs to be an SFTP server or a network drive, because drive-based and SFTP storages use a hierarchical chunk directory while most cloud storages use a flat one.  You'll need to "flatten" the chunk directory first if you want to copy from local to cloud.

lowne:
I'm a bit confused; according to https://github.com/gilbertchen/duplicacy/pull/181#issuecomment-328654782 and my own testing,
>  the chunkID is unique for each storage, even when added with the -copy to make them compatible..

at least for encrypted storages.
[...]

PR #211, opened 2017-09-25
default to single-dir-nesting for local and SFTP storages

lowne:
As per https://github.com/gilbertchen/duplicacy/issues/163#issuecomment-331242178

> [...] listing (eventually) 65k dirs over SFTP, with a full roundtrip for each one, would just take forever (it's already annoying enough within the LAN).
Single level nesting would result in ~1k files per dir with a 1TB repo with the default chunk size which hardly seems overwhelming - and would cut down dirlist requests to 256, or about 5 minutes, which is far more reasonable.

Quick and informal testing (with a tiny repo) shows that `backup`, `copy`, `prune`, `list` and `check` keep working as expected, i.e. single-nested (new) chunks can coexist alongside preexisting double-nested chunks.

### DIFF
--- a/src/duplicacy_dropboxstorage.go
+++ b/src/duplicacy_dropboxstorage.go
@@ -15,12 +15,13 @@ import (
 type DropboxStorage struct {
 	RateLimitedStorage
 
-	clients    []*dropbox.Files
-	storageDir string
+	clients         []*dropbox.Files
+	minimumNesting  int  // The minimum level of directories to dive into before searching for the chunk file.
+	storageDir      string
 }
 
 // CreateDropboxStorage creates a dropbox storage object.
-func CreateDropboxStorage(accessToken string, storageDir string, threads int) (storage *DropboxStorage, err error) {
+func CreateDropboxStorage(accessToken string, storageDir string, minimumNesting int, threads int) (storage *DropboxStorage, err error) {
 
 	var clients []*dropbox.Files
 	for i := 0; i < threads; i++ {
@@ -37,8 +38,9 @@ func CreateDropboxStorage(accessToken string, storageDir string, threads int) (s
 	}
 
 	storage = &DropboxStorage{
-		clients:    clients,
-		storageDir: storageDir,
+		clients:         clients,
+		storageDir:      storageDir,
+		minimumNesting:  minimumNesting,
 	}
 
 	err = storage.CreateDirectory(0, "")
@@ -189,11 +191,8 @@ func (storage *DropboxStorage) FindChunk(threadIndex int, chunkID string, isFoss
 		suffix = ".fsl"
 	}
 
-	// The minimum level of directories to dive into before searching for the chunk file.
-	minimumLevel := 1
-
 	for level := 0; level*2 < len(chunkID); level++ {
-		if level >= minimumLevel {
+		if level >= storage.minimumNesting {
 			filePath = path.Join(dir, chunkID[2*level:]) + suffix
 			var size int64
 			exist, _, size, err = storage.GetFileInfo(threadIndex, filePath)
@@ -217,7 +216,7 @@ func (storage *DropboxStorage) FindChunk(threadIndex int, chunkID string, isFoss
 			continue
 		}
 
-		if level < minimumLevel {
+		if level < storage.minimumNesting {
 			// Create the subdirectory if it doesn't exist.
 			err = storage.CreateDirectory(threadIndex, subDir)
 			if err != nil {
--- a/src/duplicacy_filestorage.go
+++ b/src/duplicacy_filestorage.go
@@ -18,14 +18,14 @@ import (
 type FileStorage struct {
 	RateLimitedStorage
 
-	minimumLevel    int  // The minimum level of directories to dive into before searching for the chunk file.
+	minimumNesting  int  // The minimum level of directories to dive into before searching for the chunk file.
 	isCacheNeeded   bool // Network storages require caching
 	storageDir      string
 	numberOfThreads int
 }
 
 // CreateFileStorage creates a file storage.
-func CreateFileStorage(storageDir string, minimumLevel int, isCacheNeeded bool, threads int) (storage *FileStorage, err error) {
+func CreateFileStorage(storageDir string, minimumNesting int, isCacheNeeded bool, threads int) (storage *FileStorage, err error) {
 
 	var stat os.FileInfo
 
@@ -51,7 +51,7 @@ func CreateFileStorage(storageDir string, minimumLevel int, isCacheNeeded bool, 
 
 	storage = &FileStorage{
 		storageDir:      storageDir,
-		minimumLevel:    minimumLevel,
+		minimumNesting:  minimumNesting,
 		isCacheNeeded:   isCacheNeeded,
 		numberOfThreads: threads,
 	}
@@ -137,7 +137,7 @@ func (storage *FileStorage) FindChunk(threadIndex int, chunkID string, isFossil 
 	}
 
 	for level := 0; level*2 < len(chunkID); level++ {
-		if level >= storage.minimumLevel {
+		if level >= storage.minimumNesting {
 			filePath = path.Join(dir, chunkID[2*level:]) + suffix
 			// Use Lstat() instead of Stat() since 1) Stat() doesn't work for deduplicated disks on Windows and 2) there isn't
 			// really a need to follow the link if filePath is a link.
@@ -159,7 +159,7 @@ func (storage *FileStorage) FindChunk(threadIndex int, chunkID string, isFossil 
 			continue
 		}
 
-		if level < storage.minimumLevel {
+		if level < storage.minimumNesting {
 			// Create the subdirectory if it doesn't exist.
 
 			if err == nil && !stat.IsDir() {
--- a/src/duplicacy_sftpstorage.go
+++ b/src/duplicacy_sftpstorage.go
@@ -22,12 +22,13 @@ type SFTPStorage struct {
 	RateLimitedStorage
 
 	client          *sftp.Client
+	minimumNesting  int  // The minimum level of directories to dive into before searching for the chunk file.
 	storageDir      string
 	numberOfThreads int
 }
 
 func CreateSFTPStorageWithPassword(server string, port int, username string, storageDir string,
-	password string, threads int) (storage *SFTPStorage, err error) {
+	minimumNesting int, password string, threads int) (storage *SFTPStorage, err error) {
 
 	authMethods := []ssh.AuthMethod{ssh.Password(password)}
 
@@ -36,10 +37,10 @@ func CreateSFTPStorageWithPassword(server string, port int, username string, sto
 		return nil
 	}
 
-	return CreateSFTPStorage(server, port, username, storageDir, authMethods, hostKeyCallback, threads)
+	return CreateSFTPStorage(server, port, username, storageDir, minimumNesting, authMethods, hostKeyCallback, threads)
 }
 
-func CreateSFTPStorage(server string, port int, username string, storageDir string,
+func CreateSFTPStorage(server string, port int, username string, storageDir string, minimumNesting int,
 	authMethods []ssh.AuthMethod,
 	hostKeyCallback func(hostname string, remote net.Addr,
 		key ssh.PublicKey) error, threads int) (storage *SFTPStorage, err error) {
@@ -82,6 +83,7 @@ func CreateSFTPStorage(server string, port int, username string, storageDir stri
 	storage = &SFTPStorage{
 		client:          client,
 		storageDir:      storageDir,
+		minimumNesting:  minimumNesting,
 		numberOfThreads: threads,
 	}
 
@@ -184,11 +186,8 @@ func (storage *SFTPStorage) FindChunk(threadIndex int, chunkID string, isFossil 
 		suffix = ".fsl"
 	}
 
-	// The minimum level of directories to dive into before searching for the chunk file.
-	minimumLevel := 2
-
 	for level := 0; level*2 < len(chunkID); level++ {
-		if level >= minimumLevel {
+		if level >= storage.minimumNesting {
 			filePath = path.Join(dir, chunkID[2*level:]) + suffix
 			if stat, err := storage.client.Stat(filePath); err == nil && !stat.IsDir() {
 				return filePath[len(storage.storageDir)+1:], true, stat.Size(), nil
@@ -205,7 +204,7 @@ func (storage *SFTPStorage) FindChunk(threadIndex int, chunkID string, isFossil 
 			continue
 		}
 
-		if level < minimumLevel {
+		if level < storage.minimumNesting {
 			// Create the subdirectory if is doesn't exist.
 
 			if err == nil && !stat.IsDir() {
@@ -225,7 +224,7 @@ func (storage *SFTPStorage) FindChunk(threadIndex int, chunkID string, isFossil 
 			continue
 		}
 
-		// Teh chunk must be under this subdirectory but it doesn't exist.
+		// The chunk must be under this subdirectory but it doesn't exist.
 		return path.Join(dir, chunkID[2*level:])[len(storage.storageDir)+1:] + suffix, false, 0, nil
 
 	}
--- a/src/duplicacy_snapshotmanager_test.go
+++ b/src/duplicacy_snapshotmanager_test.go
@@ -95,14 +95,14 @@ func createTestSnapshotManager(testDir string) *SnapshotManager {
 	os.RemoveAll(testDir)
 	os.MkdirAll(testDir, 0700)
 
-	storage, _ := CreateFileStorage(testDir, 2, false, 1)
+	storage, _ := CreateFileStorage(testDir, 1, false, 1)
 	storage.CreateDirectory(0, "chunks")
 	storage.CreateDirectory(0, "snapshots")
 	config := CreateConfig()
 	snapshotManager := CreateSnapshotManager(config, storage)
 
 	cacheDir := path.Join(testDir, "cache")
-	snapshotCache, _ := CreateFileStorage(cacheDir, 2, false, 1)
+	snapshotCache, _ := CreateFileStorage(cacheDir, 1, false, 1)
 	snapshotCache.CreateDirectory(0, "chunks")
 	snapshotCache.CreateDirectory(0, "snapshots")
 
--- a/src/duplicacy_snapshotmanager_test.go
+++ b/src/duplicacy_snapshotmanager_test.go
@@ -95,14 +95,14 @@ func createTestSnapshotManager(testDir string) *SnapshotManager {
 	os.RemoveAll(testDir)
 	os.MkdirAll(testDir, 0700)
 
-	storage, _ := CreateFileStorage(testDir, 1, false, 1)
+	storage, _ := CreateFileStorage(testDir, 2, false, 1)
 	storage.CreateDirectory(0, "chunks")
 	storage.CreateDirectory(0, "snapshots")
 	config := CreateConfig()
 	snapshotManager := CreateSnapshotManager(config, storage)
 
 	cacheDir := path.Join(testDir, "cache")
-	snapshotCache, _ := CreateFileStorage(cacheDir, 1, false, 1)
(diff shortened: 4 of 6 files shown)


tbain98:
I can't speak for any code that wasn't changed, but the changes that were made are clean and look good to me.

gilbertchen:
It looks good to me too.  However, I'm so worried about this change breaking storages that have been been setup and populated.  Maybe a safer way is to add a new prefix like `sftp1://` for single-dir-nesting without changing `sftp://`?

It may be ok to change `dropbox://` -- I don't think it is being used by many users.

For `file://` and `samba://`, I don't think they are nearly as slow as SFTP with two minimum levels, so maybe we don't need to change them for now?

lowne:
`dropbox://` nesting doesn't change, it was already set to 1 (inside `duplicacy_dropboxstorage.go`); I've just parametrised it for consistency between non-flat destinations.
Local and LAN storages were probably fast enough, but still double nesting causes unnecessary overhead; at the filesystem level the atomic operation is always `ls` (i.e. `ioutil.ReadDir`), and having 256 of those is strictly faster than 65k. And in general I don't think anyone would want to deal with 65k dirs with ~2 files each rather than 256 dirs with ~500 files each (I'm talking for example about tasks other than duplicacy that one might want to run on the them, such as `rsync`s, `md5sum`s, `diff`s or what have you).

So the big question is: is it safe? AFAICT the only entry point for getting chunk names (new and existing both) is `FindChunks`, and the logic there seems sound:
1. drill down until `minimumNesting` is reached
2. is the chunk there? If so, return found
3. is the next nesting subdir there? if so, drill down and goto 2
4. return not found at the current nesting level

So, if a previously-2-nested repo has `11/aa/` and `11/cc/`:
- `11aa00..` and `11cc00..` will correctly be found (or created) as `11/aa/00..` or `11/cc/00..`
- the new `11bb00` will be created as `11/bb00..`

In fact that's the exact mechanism that allows `ListAllFiles` to increase nesting when it finds a dir with too many files (i.e. no renaming needed, merely creating the subdirs ensures future chunks will be nested). So simply *decreasing* `minimumNesting` should definitely be backwards compatible, whereas increasing it would also necessitate changes in `FindChunks` to remain backwards compatible.

And lastly, I've been running this branch in my (admittedly tiny, so far) "production" environment without issues. (I've since de-nested both my local and originally-rsynced remote `chunks`, so that I could finally use the remote at all).

Having said all this, I'd still prefer to have someone else build the branch and report successful operation on existing repos before this gets merged, because messing with the repo structure is indeed inherently dangerous and it's better to be extra safe.

lowne:
To address more specifically the `sftp1://` idea: that would indeed allow "voluntary" migration from 2- to 1-nesting, and thus be deemed safer in some way, but it would also create an all-too-easy possibility for *certain* disaster, because of the problem I mentioned above with *increasing* `minimumNesting`:
- I create my repo and scripts with `sftp://`
- Oh, cool, there's single-nesting now! I update my scripts to `sftp1://`
- Weeks later, I've completely forgotten about it, and run something from the command line with `sftp://` - boom, missing chunks galore

**Ideally**, there'd be a cmdline argument for `minimumNesting` (more consistent than separate schemas imho), but dealing with the situation above (where it's *higher* than the repo's structure) requires *at a minimum* to adjust `FindChunk` (to avoid missing the lower-nested chunks that are there). In short, larger changes, therefore larger surface for potential breakage. I *had* considered it, but (at this stage at least) it's not worth it imo. Just having the (non-configurable) "default" to 1-nesting is the right thing for the vast majority of use cases.

dgcom:
Will this change allow copying one destination to another without restrictions? Ex. local to B2?
If so, it would be great - we can either use duplicacy copy or rclone sync local to B2 and then use rclone check which would use remote checksums to verify against local files without downloading everything...

lowne:
I realise I hijacked that thread (#163) with this unrelated issue, sorry everyone. @dgcom the answer you seek is at https://github.com/gilbertchen/duplicacy/issues/163#issuecomment-329527407

dgcom:
Thank you, @lowne, I saw that comment before. 
This thread is sort of logical continuation - looks like there are two conditions which needs to be met in order to bit-copy destination - it has to have _same hierarchy_ and config folder has to be copied as well....
So, I am all for making cloud destination and local file storage using same single directory approach.

lowne:
@gilbertchen I just realised that this is indeed a "breaking" change in the following scenario:
- the user has a distributed system with multiple `duplicacy` clients operating on one or more storages
- the user updates one client (to a hypothetical 2.0.10 with 1-nesting) and does a backup
- the user performs some operation with another, *outdated* client on the snapshot that has been backed up with the updated one
- missing chunks

However, I think that simply disclaiming on the release notes that updating all the clients is necessary would be sufficient, keeping in mind that
- the problematic occurrence would be quite unlikely
- even if it happens, the user would probably realise on their own that they need to keep all clients updated
- or seek support and be told so
- all without data loss - after updating the client, the chunks will be found
  - `duplicacy prune -exhaustive` is nesting-agnostic, and even deals correctly with redundant chunks that might be created by a *new* backup from the *outdated* 2-nesting version over the 1-nesting one

Additionally, after looking around more in `SnapshotManager` I'm more confident that the only change needed to allow for a generalised `-nesting 0` switch as per @dgcom (and generally changing minimum nesting back and forth at will) is indeed in `FindChunk`:
- should *disregard* `minimumNesting` (i.e. start at 0) to seek an *existing* chunk
- should *enforce* `minimumNesting` when returning the location for a *new* chunk
- slight overhead penalty

Thoughts?

tbain98:
I don't have enough time at the moment to fully read and respond to all of
the discussions from the past two days (I'll try to find that time this
weekend), but for right now I'd just like to say that I don't think we
should take @gilbertchen's suggestion that we define new pseudo-protocols
such as sftp1. I understand where that suggestion is coming from, but I
think that particular solution is one we'd kick ourselves for down the
road. The transport protocol is SFTP, so leave that alone. The choice of
how to nest the content, if baked into the URI at all, is a parameter (e.g.
?nestingLevel=1 or ?useSingleNestingLevel=true), not a protocol.

gilbertchen:
I would like to propose the following changes:

* Adding a new parameter `nesting-level` with a fixed value of 1 to the `config` file
* Adding another new parameter `adaptive-nesting` that can be `true` or `false` to the `config` file
* Adding a new option to the `init` command to set the `adaptive-nesting` (but `nesting-level` can't be changed)
* If the `config` file does not have these two parameters, all storage backends work the same way as before (for local/sftp nesting-level=2 and adaptive-nesting=true, for others nesting-level=0 and adaptive-nesting=false)
* If `nesting-level=1` is present, all storage backends including all those cloud ones will start at a nesting level of 1
* Only local/sftp/dropbox will honor `adaptive-nesting`; cloud storages won't do adaptive-nesting at all.

The reasons there are two kinds of chunk structures are:
1. the number of files per directory on Linux is limited according to https://stackoverflow.com/questions/466521/how-many-files-can-i-put-in-a-directory, so it seemed adaptive-nesting is the solution
2. But for cloud storage the adaptive nesting means multiple GetFileInfo calls per chunk, and I think this should be generally avoided

Re-examining 1, only ext2 has problems dealing with more than 10,000 files per directory; ext3 has a much larger limit and it is unlimited for ext4.  Therefore it may make sense to let user decide whether to do adaptive nesting or not.  An ext2 storage with a nesting level of 1 and no adaptive nesting can have a capacity of 4MB * 10,000 * 256 = 10TB before running into problems and I think most users won't hit that limit.

Therefore I think adaptive nesting should be turned off by default.  And then by requiring a nesting level of 1 for all backends we ensure the interoperability with rsync/rclone.  A nesting level of 1 also makes it easier to browse the chunk directory on cloud providers' websites.

For this PR I think we should just keep the parameterization but retain the old values for the nesting level.

tbain98:
A couple of things:

1. I haven't looked at the code in question, but what Mark wrote about only
considering minimumNesting when creating a chunk was exactly what I would
have expected. We should implement that algorithm. This addresses rsync
compatibility better than simply requiring all backends to use the same
configuration.
2. Don't add a parameter to the config file if you don't intend to respect
a non-default setting the user enters there. So if you plan not to allow
the user to specify minimumNesting, don't include it in the config file.
3. I disagree that adaptiveNesting should be false by default. Rather, we
should ensure that -copy respects the target storage's settings and writes
the content to a path that's compatible with them, and users who want to
use non-Duplicacy methods of copying files between storages that use
different adaptiveNesting values (which is a small fraction of the total
Duplicacy users) can disable it.

lowne:
So I think we're all more or less agreeing on these vague design goals:
1. file storages' nesting **must** be 1 at minimum and 2 at maximum (anything beyond that is in the PB realm and wouldn't interface with "consumer" filesystems anyway), 
2. cloud storages' nesting level **must** be 0 to 2
3. users **must** be able to `duplicacy copy` regardless of nesting schemas
4. (advanced) users **must** be able to `rsync/rclone`/etc via non-trivial scripting for navigating incompatible nesting
5. duplicacy **must** keep working with existing storages
6. file storages' nesting **should** default to 1 for reasonable efficiency
7. file storages' nesting **should** be adaptive as they grow
8. cloud storages' nesting **should** be fixed to minimise lookups
9. (advanced) users **should** be able to `rsync/rclone`/etc given some provisos, config flags, etc
10. (advanced) users **should** be able to `rsync/rclone`/etc transparently
11. any new schema **should** allow users to (at least manually) migrate existing storages
12. any new schema **should** support existing storages transparently and automatically

Obviously the status quo respects all the musts 1-5 plus 7 and 8. ( @tbain98 for 3, `ChunkUploader` also goes through the storage's `FindChunk`, which is where the nesting rules are encoded.)

My original solution only focuses on 6 and satisfies 7, 8, 11, 12, while expanding the scope to cloud storages (9 and 10) poses a challenge.

An immediate contradiction arises between 8 and 12; @gilbertchen's solution ditches 12 (and 10 and 11) in favour of maintaining 8 and adding 9, and this effectively introduces a `version` flag to the storage with distinct algorithms; the `nesting-level=1` is basically just a version flag in disguise.

There are several solutions for the alternative way of sacrificing 8 for the sake of 12/11/10; to minimise the impact of cloud lookups I suggest the following that generalises on `nesting-level`:
- I agree that to greatly simplify things, we want to converge all storages to single-nesting without adaptive
- the storage's `config` adds a flag in the form `nesting-level=U[,S1[,S2]]` (all of which can be 0, 1 or 2)
- `FindChunk` should be abstracted away (in StorageManager) and behave as such:
  - if `chunkID` exists at level `U`, return found at level `U`
  - elseif `S1` and exists at level `S1`, return found at level `S1`
  - elseif `S2` and exists at level `S2`, return found at level `S2`
  - else return not found at level `U`
  - basically S1 and S2 are "adaptive"/compatibility switches (objectives 11, 12)
  - when both are omitted, nesting is fixed (objectives 8, 10)
- `config` of existing storages gains on first access (objective 5):
  - `nesting-level=2` for file storages
  - `nesting-level=0` for cloud storages
  - `nesting-level=1` for dropbox
- `config` on new storages defaults to `nesting-level=1` for all
- duplicacy prints a suggestion to run `duplicacy set nesting-level=1` on operations on file storages with level 2
- `duplicacy set nesting-level=nU[,nS1[,nS2]]` does the following:
  - set `nesting-level` to `nU,U,nS1,nS2,S1,S2` without duplicates (because of set `0..2`, there'll be 3 numbers max) and save `config`
  - get `ListAllFiles('chunks')`, including fossils
  - exclude all files that regex match nesting levels `nU`, `nS1`, `nS2`
  - rename all remaining files so they're at nesting level `nU` (requires proper rename function for cloud storages)
  - maybe an extra `ListAllFiles` to make double-sure that there are no remaining chunks or extraneous files outside of `nU`, `nS1` or `nS2`
  - if everything is a-ok, finally set `nesting-level=nU,nS1,nS2` and save `config`
  - this should guarantee (unless I'm overlooking something) that `set nesting-level` is idempotent, restorable, and doesn't break things even if it fails for any reason
  - the possible exception is for fossil collections going out of sync? (idk the internal details, i.e. if they store filenames or just IDs and rely on `FindChunk` like everything else) (an easy cop out could be to ignore it and maybe suggest a `prune -exhaustive`)

Usage cases:
- user just wants to quickly stop the proliferation of further subdirs on local storage: `duplicacy set -storage local -nesting-level=1,2`
- user wants to flatten and make trivially-`rsync`-compatible their storages: `duplicacy set -storage ... -nesting-level=1`
- as above, but storages have now grown too much and user wants to manually start nesting further without touching the existing chunks (i.e. "adaptive nesting" of sorts): `duplicacy set -storage ... -nesting-level=2,1`
- user starts a new storage >100TB (or tiny chunks) on ext2: `duplicacy init ... -nesting-level=2`

The only case where `S2` is at all needed is with "super-adaptive" 0/1/2-nesting storages, which seems indeed rare - so there's a case to just not bother with it.

gilbertchen:
Here is my new proposal, taking some of the ideas suggested by @lowne:

* Don't change the `config` file; instead, add a new file perhaps called `nesting` that defines the nesting levels
* This `nesting` file is a json file that has two keys: `read-levels` and `write-level`
* The value of `read-levels` is an array of integers representing which levels the chunks should be read from.  For example, `"read-levels": [0, 1, 2]` means chunks must be searched in level 0, level 1, level 2, in that order
* The value of `write-level` is an integer representing the level that chunks should be written to.  For example `"write-level": 1` means chunks will be written to level 1.
* In the new version, all storage backends will respect the levels defined in this file.  If this file is not present or if it is an old version of Duplicacy it will behave the same way as before.
* This file can be freely edited by users.  A simple command can be implemented to edit this file but I don't think it is necessary.

I think this solution not only ensures rsync/rclone interoperability, but also gives the ultimate freedom to users on how to organize the chunk directory.  It is better that adaptive nesting for two reasons:
* Most cloud storages can't support adaptive nesting because they don't have the notion of directories so they can't check if the directory for the next level exists
* This is more efficient because the number of lookups is equal to the number of read levels, while in adaptive nesting you needs extra calls for detecting if directories exist.

tbain98:
I've spent the past few days thinking about exactly how to respond to both
@lowne's and @gilbertchen's proposals. A part of the trouble I had was that
nowhere that I've seen have we actually stated what problem we're trying to
solve (so I'm going to do that to make sure we're all working towards the
same goals). Another part was I disagreed with some parts of both solutions
(though not necessarily the broad goal). And finally, I think that all the
ideas we're kicking around might be out of scope for the PR we're attaching
them to and therefore delaying the incorporation of a positive (though not
complete) improvement into the code, so I propose that we open a new issue
for the big-picture enhancements and leave this PR tightly focused on
defaulting file storages to level 1.

With that said, I'm currently on a plane without access to GitHub, so I can
use email to add comments to this PR but can't submit a new issue, so I'm
going to do that for now. But I do think we should open a new issue.

So first, what are we trying to do here? I believe that we're trying to
implement a new feature to allow simple copying or comparing of content
between any pair of storages via external tools such as rsync or rclone.
This isn't a requirement I personally am interested in (nor even am sure
would be possible for most of the cloud storage providers, because 1. I'm
not sure rsync and friends would work against them, and 2. if many of them
don't support directories per @gilbertchen's bullet near the bottom, then
how could you sync with a file-based storage that uses nesting?), but I'm
willing to accept on faith that it's a real requirement for certain users
even if not for me. The problem with meeting that requirement in the
current implementation is that different storage providers have different
optimal configurations for how many directories to nest. Also, the adaptive
growth algorithm as currently implemented is problematic in its own right
because it requires more lookups to determine whether a chunk exists
(because it could be at multiple different levels). Adaptive nesting as
currently implemented is also a problem when using external tools to sync
two storages, because the same chunk might exist at different depths in the
two storages and might be copied in both directions, resulting in duplicate
chunks and the possibility of having more chunks in a higher-level
directory than the configured limit. Did I miss anything?

OK, on to the concerns I have about what the two of you are proposing.

1. I'm opposed to the idea of making Duplicacy unconditionally less
efficient (in terms of run time or cost of API calls) and/or excluding
storage providers that don't support directories simply to "enable"
rsyncing between storage providers who can't even support those tools. I'm
not in favor of the one-size-fits-all approaches @lowne proposed, and his
statement that we're all in agreement that we should converge on
non-adaptive fixed nesting is incorrect. But I'm not at all opposed to
having the ability to choose to do that via configuration; I don't mind
features I don't want that carry significant drawbacks as long as I'm not
forced to use them. I think that @gilbertchen's proposal allows users to
not use his proposed feature by simply not setting the nesting settings at
a global level (right?), so I don't object to it on that front.
2. I'm not seeing the benefit of explicitly specifying the search order of
the levels when reading. Why wouldn't we start at the deepest level (which
has the ability to hold the most files and therefore has the greatest
chance of holding a particular file) and then work upward if not found?
3. I'm not a fan of adding a warning message to nag the user about their
choice of configuration if they choose it intentionally. I'm fine with it
if they haven't taken a conscious action to use that setting (i.e. if they
just upgraded to a version that has that setting), but if they made the
choice consciously, Duplicacy should shut up and do what the user asked
for, and @lowne's proposal doesn't do that. This implies that Duplicacy
shouldn't modify your config file for you (so you can tell a conscious
choice from inaction), which is probably the right choice anyway (software
really shouldn't be unilaterally modifying its own config files).
4. I like @lowne's idea of an algorithm for moving chunks to a lower
directory when the nesting level changes, but a) I'm not sure it's
thread-safe in the face of a currently-running backup from another host,
and b) we should consider applying it automatically when doing adaptive
nesting in order to allow the flexibility of adaptive nesting but get the
performance improvement of only having to search a single directory.
5. My immediate reaction to @gilbertchen's proposal that we make a new
config file for nesting was "Dear God, why?" If you have a reason for
having it not be part of the main config file, I'm curious to hear what it
is (reuse?), but if it's "because the config isn't JSON," my response is
"then let's make the config file JSON."
6. I think that when a global setting like @gilbertchen's is not in use,
each individual storage should have a configuration for what level to begin
writing at, how many files in a single directory should trigger the
creation of the next level down (set to - 1 for fixed nesting, including
cloud storages), a flag for whether to write at the lowest level for all
new writes (otherwise you deepen each individual directory only when it
itself is full), and a flag for whether to migrate files downward when
expanding the directory tree or to leave already-written files in place (so
you have to search parent directories). I think these changes can co-exist
with the changes @gilbertchen has proposed; it's not an either-or. One
allows a global configuration, while the other allows configuration of
individual storages to allow them each to use optimal configuration.

gilbertchen:
I agree that the discussion is far beyond the scope of this PR.  I'll create a new issue presenting my solution and looking for feedback.

For this PR, however, I'm still concerned about breaking backward compatibility with nesting level changes.  @lowne can we have most of your code changes but revert all the levels to their original values?  I would be more comfortable if the nesting level can be explicitly changed, like the `nesting` file that I proposed.

lowne:
@gilbertchen done. 
FWIW I think @gilbertchen's latest proposal is satisfactory (the main non-ideal aspect is that the nesting *should* be inside the `config` file imho, but that means that changes need to go through a `duplicacy` command, which then should also probably check for consistency... complications arise, so avoiding those and delegating to the user's care is not the end of the world).

lowne:
@tbain98 - maybe you missed my expanded & clarified update due to email interface; as for your points:
1. bad wording on my part - I meant that *I* agree (with @gilbertchen) wrt no adaptive - and I stand by that: for 8-bit nesting (as opposed to 4-bit), adaptive either wouldn't ever occur (for the majority of people), or its benefits are questionable (because people with big data sets tend to use proper filesystems), or it's something that can only happen *once* anyway (and my solution allows for it).
2. to minimise lookup time; but in practice my `S2` (or @gilbertchen's 3rd seek value) would never be used anyway
3. agreed, that was my intention as well (admittedly I didn't go over every detail)
4. it's thread-safe with the 2nd `ListAllFiles` check as it'd leave the storage in the backwards-compatible state (and it's restorable)
5. as noted above, I agree, but the `config` file is not (and should not be) user-editable
6. unless I'm completely wrong, everything discussed is about individual storages (i.e. backup destinations)

lowne:
Uh, I just found @gilbertchen's proper discussion thread at #222 - apologies for continuing the discussion here, but I hadn't seen it linked so I assumed it wasn't created yet. (Also for anyone else who might want to further the general discussion: go to #222)

gilbertchen:
Thank you for your contribution.  I'll be working on #222 next week.

Issue #206, opened 2017-09-21
Fully document build environment used to produce releases.


macdanny:
There is a significant difference in "duplicacy copy" performance between the releases on GitHub and builds I produce myself.

Using the GitHub release, my "duplicacy copy" command averages around 7.25 MB/s.

If I produce a build with "go get -u github.com/gilbertchen/duplicacy/duplicacy" and use the resulting binary, my "duplicacy copy" command averages only around 4 MB/s.

This is an issue because "duplicacy copy" frequently crashes and I have to start over. Code was introduced after 2.0.9 that improves the performance when starting "duplicacy copy" over again. I have to choose either good download performance or a reasonable chance of completion. I would like to have both.

I don't believe this is an issue with the duplicacy code but rather with one of the libraries it uses. On my machine I checked out and built v2.0.9 and built it and the performance is still poor.

go version go1.8.3 linux/amd64

These are my revisions.

<pre>
./golang/protobuf
commit 130e6b02ab059e7b717a096f397c5b60111cae74
./kr/fs
commit 2788f0dbd16903de03cb8186e5c7d97b69ad387b
./satori/uuid
commit 5bf94b69c6b68ee1b541973bb8e1144db23a194b
./dgrijalva/jwt-go
commit a539ee1a749a2b895533f979515ac7e6e0f5b650
./aryann/difflib
commit e206f873d14a916d3d26c40ab667bca123f365a3
./Azure/go-autorest
commit f6be1abbb5abd0517522f850dd785990d373da7e
./googleapis/gax-go
commit 317e0006254c44a0ac427cc52a0e083ff0b9622f
./gilbertchen/azure-sdk-for-go
commit 0b09de4174ca0cadcd3abb4aa31267c6f5ccebbb
./gilbertchen/go-dropbox
commit 90711b603312b1f973f3a5da3793ac4f1e5c2f2a
./gilbertchen/goamz
commit eada9f4e8cc2a45db775dee08a2c37597ce4760a
./gilbertchen/gopass
commit bf9dde6d0d2c004a008c27aaee91170c786f6db8
./gilbertchen/keyring
commit 1260032d8c5455033c8aa592564d8ea0a6874935
./gilbertchen/duplicacy
commit 978212fd756a8946e203b34168ec027793260ddb
./gilbertchen/cli
commit 1de0a1836ce9c3ae1bf737a0869c4f04f28a7f98
./gilbertchen/xattr
[...]

PR #214, opened 2017-09-26
Retry downloads with corrupted content up to three times.

macdanny:
Wasabi's GetObject occasionally (approximately 2% of the time in my testing) returns objects whose contents disagree with what has been stored in Wasabi. These cause errors when chunks are downloaded (during restore, for example). Previously, these errors would abort the restore, requiring that it be started over from the beginning. This made it effectively impossible to complete any normally-sized restore where the cumulative chance of encountering such an error approaches unity.

With this change Duplicacy will retry up to three times if it can't decrypt the downloaded chunk, or if the downloaded chunk's ID doesn't agree with a chunk ID computed from the downloaded chunk's content.

### DIFF
--- a/src/duplicacy_chunkdownloader.go
+++ b/src/duplicacy_chunkdownloader.go
@@ -324,22 +324,39 @@ func (downloader *ChunkDownloader) Download(threadIndex int, task ChunkDownloadT
 		LOG_DEBUG("CHUNK_FOSSIL", "Chunk %s has been marked as a fossil", chunkID)
 	}
 
-	err = downloader.storage.DownloadFile(threadIndex, chunkPath, chunk)
-	if err != nil {
-		LOG_ERROR("UPLOAD_FATAL", "Failed to download the chunk %s: %v", chunkID, err)
-		return false
-	}
+	const MaxDownloadAttempts = 3
+	for downloadAttempt := 0;; downloadAttempt++ {
+		err = downloader.storage.DownloadFile(threadIndex, chunkPath, chunk)
+		if err != nil {
+			LOG_ERROR("UPLOAD_FATAL", "Failed to download the chunk %s: %v", chunkID, err)
+			return false
+		}
 
-	err = chunk.Decrypt(downloader.config.ChunkKey, task.chunkHash)
-	if err != nil {
-		LOG_ERROR("UPLOAD_CHUNK", "Failed to decrypt the chunk %s: %v", chunkID, err)
-		return false
-	}
+		err = chunk.Decrypt(downloader.config.ChunkKey, task.chunkHash)
+		if err != nil {
+			if downloadAttempt < MaxDownloadAttempts {
+				LOG_WARN("RETRY_DOWNLOAD", "Failed to decrypt the chunk %s: %v", chunkID, err)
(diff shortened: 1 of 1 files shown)


src/duplicacy_chunkdownloader.go
```diff
@@ -324,22 +324,39 @@ func (downloader *ChunkDownloader) Download(threadIndex int, task ChunkDownloadT
 		LOG_DEBUG("CHUNK_FOSSIL", "Chunk %s has been marked as a fossil", chunkID)
 	}
 
-	err = downloader.storage.DownloadFile(threadIndex, chunkPath, chunk)
-	if err != nil {
-		LOG_ERROR("UPLOAD_FATAL", "Failed to download the chunk %s: %v", chunkID, err)
-		return false
-	}
+	const MaxDownloadAttempts = 3
+	for downloadAttempt := 0;; downloadAttempt++ {
```
tbain98: This loop would be cleaner if you pull out a helper method that attempts to download, decrypt, and compare the chunk's ID, and returns true (plus any other needed return values, though I'm not seeing any from a quick look) if all those things succeeded and false if any of them failed. Then your loop's body could handle the logic of what to do if you've failed (checking if you're past the limit or not) in a single place.

That separates the concerns (attempting a single download vs. retrying 3 times), and also starts to carve up the 130-line method into something more reasonably-sized.

gilbertchen:
Thank you for your work.  This is a very nice feature.

I haven't heard back from Wasabi.

tbain98:
@gilbertchen I'm curious: both you and I made comments about small changes that would improve this code, which @macdanny hadn't yet implemented, but you merged the PR without those fixes. What was your thinking? Did you just decide that the improvements were small enough to not be worth waiting for?

gilbertchen:
@tbain98 it is easier for me to merge the PR first and then make the changes myself: https://github.com/gilbertchen/duplicacy/commit/b2ad6da36477328344caac09afb77edb5688bae7

You can also make the changes yourself and submit a PR.  I'm going to vendor the `src` directory for #206 and wanted to empty the PR queue as much as possible before doing that.

tbain98:
Ah, makes sense. Thanks for the explanation.

macdanny:
Sorry I haven't been very responsive, it's hard to find time for side projects sometimes. I have today off and I was going to look at this again.

@tbain98 I agree that the code could be improved, and I wasn't opposed to making those changes. The reason I wrote the code in the way that I did was to keep the change footprint small to make it easier to review and accept. I attempted to keep the shape of the code similar to what @gilbertchen committed in 8808ad5, which is similar in purpose.

@gilbertchen Thanks for merging this. Sorry I was too slow to make the changes you requested.

I doubt Wasabi is going to be able to fix this very easily. And in fact maybe it's not all that important that they fix it. I did the math wrong in my calculation, it didn't impact 2% of chunks, but 0.2% - 2 per thousand. While I think that's pretty bad, there's no way that number is ever going to go to zero. They provide mechanisms that can be used to be tolerant of the inevitable transmission failures and at the end of the day, taking advantage of those mechanisms is probably the right thing to do anyway.

Issue #124, opened 2017-08-24
gdrive restore 503 problem


TheBestPessimist:
![capture](https://user-images.githubusercontent.com/4482210/29661214-05476f88-88cc-11e7-8ae0-e47ca1e5d561.PNG)

this happened after more than 2 hours of restoring  a backup with 20 threads.

gilbertchen:
Not sure about this one.  503 error is retried, so maybe the server returned something not parseable? Did the log show other errors before this one?

TheBestPessimist:
There is no log for restores :( (in gui)

TheBestPessimist:
Nothing else shown, that's why i took the screenshot

TheBestPessimist:
In CLI at least, this is fixed for me, so i am going to close it now.

Issue #123, opened 2017-08-24
bug: restoring from gdrive on gui: rate limit exceeded with 64 threads


TheBestPessimist:
I was testing the gui version (2.0.8) in a vm, trying to restore with 64 threads some 20+ GB of stuff, 30k folders, 77k files (linux kernel, photos, videos + duplicates of those -- all the general file sizes), and a few minutes after starting restore i got an error about rate limit exceeded, and the restore stopped.

"error failed to find the chuck <some chunk id here>: googleapi: error 403: user rate limit exceeded, userRateLimitExceeded."

From what i remembered this error was fixed in 2.0.6, right?

gilbertchen:
Error 403 should be retried.  Sorry, just found out the FindChunk function doesn't retry on any error at all -- will fix that.

TheBestPessimist:
waiting for the fix. will close after it's fixed and i can restore properly (and yes, need more than 30 threads to fill up my 1Gbit connection).

gilbertchen:
I was wrong.  Although FindChunk doesn't retry, it calls listByName which does retry.  So I think the issue could be that the maximum number of retries (8) had been reached and the 403 error was returned as it was.

Can you try to run the CLI version from the command line to see if this is the case?  I'll run my own tests too.

TheBestPessimist:
I ran in cmd, and i can say that the rate limiting happens only sometimes there. It's rather inconsistent. 
Beside duplicacy -v, is there any other logging option which i can give you?

TheBestPessimist:
With 64 threads i also got 403 during backup, not only restore
```
Skipped chunk 834 size 4055755, 13.65MB/s 00:07:10 40.4%
Failed to find the path for the chunk a52bba2093b347d270faf6e46bec69dca46c105acb94f63a86f3a959fa018105: googleapi: Error 403: User Rate Limit Exceeded, userRateLimitExceeded
```

TheBestPessimist:
It failed (during backup) even with 24 threads:

```
Failed to find the path for the chunk 7824eb3d4785283a7773294e4c2a3b2e122ed0bf046f5ef3b56de70fd24e4fbf: googleapi: Error 403: User Rate Limit Exceeded, userRateLimitExceeded

Z:\backup_software_test\the_data>Duplicacy_win_x64_2.0.8  -v backup -threads 24 -stats
```

gilbertchen:
'-d' instead of '-v' will show retry messages.

[...]

PR #221, opened 2017-10-01
Tbp/google rate limit exceeded

TheBestPessimist:
With this i want to fix/delay #123  and #124.

This is the implementation i have come up with regarding the backoff computation for GCD storage.

Maybe this could be made in  a standalone file, and applied somehow to all storages?

### DIFF
--- a/src/duplicacy_gcdstorage.go
+++ b/src/duplicacy_gcdstorage.go
@@ -108,16 +108,17 @@ func (storage *GCDStorage) shouldRetry(threadIndex int, err error) (bool, error)
 		return false, err
 	}
 
-	delay := storage.backoffs[threadIndex]*rand.Float64() + storage.backoffs[threadIndex]*rand.Float64()
-	LOG_INFO("GCD_RETRY", "Thread: %03d. Message: %s. Retrying after %.2f seconds. current backoff: %.2f.  Number of retries: %d", threadIndex, message, delay, storage.backoffs[threadIndex], storage.backoffsRetries[threadIndex])
-	time.Sleep(time.Duration(delay * float64(time.Second)))
-
 	if storage.backoffs[threadIndex] < LIMIT_BACKOFF_TIME {
 			storage.backoffs[threadIndex] *= 2.0
 	} else {
 		storage.backoffs[threadIndex] = LIMIT_BACKOFF_TIME
 		storage.backoffsRetries[threadIndex] += 1
 	}
+
+	delay := storage.backoffs[threadIndex]*rand.Float64() + storage.backoffs[threadIndex]*rand.Float64()
+	LOG_INFO("GCD_RETRY", "Thread: %03d. Message: %s. Retrying after %.2f seconds. Current backoff: %.2f. Number of retries: %d", threadIndex, message, delay, storage.backoffs[threadIndex], storage.backoffsRetries[threadIndex])
+	time.Sleep(time.Duration(delay * float64(time.Second)))
+
 	return true, nil
 }
 
--- a/src/duplicacy_gcdstorage.go
+++ b/src/duplicacy_gcdstorage.go
@@ -5,23 +5,24 @@
 package duplicacy
 
 import (
-    "io"
+	"io"
 	"fmt"
 	"net"
-    "path"
-    "time"
-    "sync"
-    "strings"
+	"path"
+	"time"
+	"sync"
+	"strings"
 	"net/http"
 	"net/url"
-    "io/ioutil"
-    "math/rand"
-    "encoding/json"
+	"io/ioutil"
+	"math/rand"
+	"encoding/json"
 
 	"golang.org/x/net/context"
 	"golang.org/x/oauth2"
 	"google.golang.org/api/drive/v3"
 	"google.golang.org/api/googleapi"
+	"runtime"
 )
 
 type GCDStorage struct {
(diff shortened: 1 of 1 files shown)


src/duplicacy_gcdstorage.go
```diff
@@ -27,10 +27,11 @@ import (
...
 
-	service     *drive.Service
-	idCache     map[string]string
-	idCacheLock *sync.Mutex
-	backoffs    []int
+	service         *drive.Service
+	idCache         map[string]string
+	idCacheLock     *sync.Mutex
+	backoffs        []float64
+	backoffsRetries []int
```
tbain98: Because the variable name could be interpreted ambiguously, I recommend you add a comment to say that this is the number of backoffs that have been performed so far for the current send attempt on the given thread. Also, it's a small thing, but I think the variable name would be more clear as backoffRetries (or as backoffRetriesPerThread) than as backoffsRetries.

src/duplicacy_gcdstorage.go
```diff
@@ -45,11 +46,19 @@ type GCDConfig struct {
 }
 
 func (storage *GCDStorage) shouldRetry(threadIndex int, err error) (bool, error) {
+	const LIMIT_BACKOFF_TIME = 64
```
tbain98: This whole 6-line block looks like it could be pulled out as member variables (or consts), because they don't appear to change between method calls. That would avoid spending a few CPU cycles on each method call, but also slim down the method body to focus on the actual algorithm.

src/duplicacy_gcdstorage.go
```diff
@@ -62,8 +71,9 @@ func (storage *GCDStorage) shouldRetry(threadIndex int, err error) (bool, error)
 			retry = true
 		} else if e.Code == 403 {
 			// User Rate Limit Exceeded
-			message = "User Rate Limit Exceeded"
+			message = e.Message // "User Rate Limit Exceeded"
```
tbain98: Looks like you accidentally left the previous message commented out; please remove.

src/duplicacy_gcdstorage.go
```diff
@@ -62,8 +71,9 @@ func (storage *GCDStorage) shouldRetry(threadIndex int, err error) (bool, error)
...
 			retry = true
+
```
tbain98: Superfluous and not consistent with the surrounding code; remove.

src/duplicacy_gcdstorage.go
```diff
@@ -83,18 +93,37 @@ func (storage *GCDStorage) shouldRetry(threadIndex int, err error) (bool, error)
...
+		storage.backoffs[threadIndex] = computeInitialBackoff(minimumSleep, maximumSleep)
+		storage.backoffsRetries[threadIndex] = 0
 		return false, err
 	}
 
-	delay := float32(storage.backoffs[threadIndex]) * rand.Float32()
-	LOG_DEBUG("GCD_RETRY", "%s; retrying after %.2f seconds", message, delay)
-	time.Sleep(time.Duration(float32(storage.backoffs[threadIndex]) * float32(time.Second)))
-	storage.backoffs[threadIndex] *= 2
+	if storage.backoffs[threadIndex] < LIMIT_BACKOFF_TIME {
```
tbain98: This logic is going to result in you going over your limit, which isn't the goal. You also have a bug where you're only incrementing the retries count when you're over the limit.

Instead, do the following:

```storage.backoffs[threadIndex] *= 2
if storage.backoffs[threadIndex] > LIMIT_BACKOFF_TIME {
    storage.backoffs[threadIndex] = LIMIT_BACKOFF_TIME
}
storage.backoffsRetries[threadIndex]++
```

src/duplicacy_gcdstorage.go
```diff
@@ -83,18 +93,37 @@ func (storage *GCDStorage) shouldRetry(threadIndex int, err error) (bool, error)
...
+		storage.backoffs[threadIndex] *= 2.0
+	} else {
+		storage.backoffs[threadIndex] = LIMIT_BACKOFF_TIME
+		storage.backoffsRetries[threadIndex] += 1
+	}
+	delay := storage.backoffs[threadIndex]*rand.Float64() + storage.backoffs[threadIndex]*rand.Float64()
```
tbain98: This merits an explanatory comment.

src/duplicacy_gcdstorage.go
```diff
@@ -45,11 +46,19 @@ type GCDConfig struct {
 }
 
 func (storage *GCDStorage) shouldRetry(threadIndex int, err error) (bool, error) {
+	const LIMIT_BACKOFF_TIME = 64
+	const MAX_NUMBER_OF_RETRIES = 15
+	minimumSleepRatio := 0.1
+	maximumSleepRatio := 0.2
+	minimumSleep := float64(storage.numberOfThreads) * minimumSleepRatio
```
tbain98: How about a comment explaining why you're using this algorithm (threads * 0.1 or 0.2) and what this particular algorithm is trying to accomplish. As-is, a casual reviewer has no way to know why we're doing this nor whether it's the right thing to be doing.

src/duplicacy_gcdstorage.go
```diff
@@ -83,18 +93,37 @@ func (storage *GCDStorage) shouldRetry(threadIndex int, err error) (bool, error)
...
+		storage.backoffsRetries[threadIndex] += 1
+	}
+	delay := storage.backoffs[threadIndex]*rand.Float64() + storage.backoffs[threadIndex]*rand.Float64()
+	LOG_DEBUG("GCD_RETRY", "Thread: %3d. Message: %s. Retrying after %6.2f seconds. Current backoff: %6.2f. Number of retries: %2d.", threadIndex, message, delay, storage.backoffs[threadIndex], storage.backoffsRetries[threadIndex])
+	time.Sleep(time.Duration(delay * float64(time.Second)))
+
 	return true, nil
 }
 
+/*
```
tbain98: Although I applaud the explanatory comments, I can't help feeling that most of this content is in the wrong place. In this method you've implemented an algorithm to choose a random number between min and max, but 1) there's nothing in this method about 0.1 and 0.2, that part of the comment belongs on the code that uses those values, 2) although the SO reference could be useful, you'd be better off just leaving it off and writing in plain English that you're picking a random number between min (inclusive) and max (inclusive), 3) it's not at all clear why Google won't be DDOS'ed because of the choice to sleep between min and max (or is this still part of the commentary on threads * 0.1 or 0.2, covered by my comment 1?).

src/duplicacy_gcdstorage.go
```diff
@@ -274,7 +303,12 @@ func CreateGCDStorage(tokenFile string, storagePath string, threads int) (storag
 		numberOfThreads: threads,
 		idCache:         make(map[string]string),
 		idCacheLock:     &sync.Mutex{},
-		backoffs:        make([]int, threads),
+		backoffs:        make([]float64, threads),
+		backoffsRetries: make([]int, threads),
+	}
+
+	for b := range storage.backoffs {
+		storage.backoffs[b] = 0.1 * float64(storage.numberOfThreads) // at the first error, we should still sleep some amount
```
tbain98: Why isn't this calling `computeInitialBackoff()`?

src/duplicacy_gcdstorage.go
```diff
@@ -45,11 +46,19 @@ type GCDConfig struct {
 }
 
 func (storage *GCDStorage) shouldRetry(threadIndex int, err error) (bool, error) {
+	const LIMIT_BACKOFF_TIME = 64
+	const MAX_NUMBER_OF_RETRIES = 15
+	minimumSleepRatio := 0.1
+	maximumSleepRatio := 0.2
+	minimumSleep := float64(storage.numberOfThreads) * minimumSleepRatio
+	maximumSleep := float64(storage.numberOfThreads) * maximumSleepRatio
+	rand.Seed(time.Now().UnixNano()) // unsure if this is needed
```
tbain98: It's needed once per program execution, but not once per method call, otherwise you're just wasting CPU cycled re-initializing it each time. It should be called near the top of `main()` if it's not already being called there, and then not called here.

TheBestPessimist:
Will work on the changes in a few days. Right now I'm on a 15 hours plane ride. Bummer.

gilbertchen:
I ran a few tests with this PR and it has been working great -- with 64 threads and a very small chunk size (1k) it kept going despite tons of 'rate exceeded' messages.  Thanks for the great work!

I'll make some minor changes after the merge, mostly in response to @tbain98's review comments.

TheBestPessimist:
Ah, i forgot about this, thought it was already merged.  Sorry for not fixing everything. If i do it tonight, could you merge that as well, or it's easy to late?

gilbertchen:
I created PR #236. Please review the changes I made there.

PR #223, opened 2017-10-03
add more detailed revision deletion logging on prune

lowne:


gilbertchen:

PR #243, opened 2017-10-21
GUIDE.md: Fix display of asterisk

pawitp:
Asterisk should be escaped otherwise it will be shown as italics.


gilbertchen:
Thank you for the fix!

Issue #137, opened 2017-08-28
Security weakness: PBKDF2 invocation uses static salt reducing effective security


shonjir:
The PBKDF2 key generation function uses a hard-coded static salt and hard-coded number of rounds.  This dramatically reduces the keyspace of the generated encryption keys. 

https://github.com/gilbertchen/duplicacy/blob/7f04a791119759e903946375a1463ca3acab3110/src/duplicacy_config.go#L25

https://github.com/gilbertchen/duplicacy/blob/22ddc04698077fc297ae7408991405fc4aac0fad/src/duplicacy_utils.go#L118

PBKDF2 is an HMAC based key derivation function.  It derives its strength from the number of rounds and through the use of a random salt to prevent rainbow table style attacks.

It is essential to use a randomly derived salt of at least 8 bytes for this function to establish sufficient levels of security for the generated keys.

Best practice for this function is to securely generate a new random salt along with a new password (e.g., when the dataset is created).  The key derivation function, salt, number of rounds, and the hash function identifier should be stored with the dataset so that the generated key may be reproduced regardless of the current key generation method.  Making the number of rounds configurable (above a minimum default security level) would further increase the security of the implementation. 

Because this key is used directly as a data encryption key the use of hard-coded salt and hard-coded rounds means that these values cannot be changed in the future without invalidating existing backup-sets.

The key derivation function should be used instead as a key-encryption-key to protect a randomly generated data-encryption-key generated when the dataset is created.  This would also the user to change the dataset passwords as well as allow later implementation of alternative protection methods for the data-encryption-key such as PKCS, and allow leveraging cryptographic key stores, etc.

gilbertchen:
I agree this is a security flaw.  A static salt makes it possible for the potential hackers to build a universal rainbow table to attack a large number of duplicacy users at once if they can, for instance, gain access to a cloud storage.

So the V2 of the master key encryption scheme should be like this:
* generate a random 8 byte (or more) salt
* set the number of rounds to 32768 (or more)
* call the PDKDF function to generate a key and use that to encrypt the config file
* add a header indicating the key derivation function, the random salt, and the number of rounds to the encrypted config file
* the init command can have a switch to increase the number of rounds
* for backward compatibility, try V1 decryption first and switch to V2 on a V1 decryption error

Do you see any issue with these?  Any suggestions?

gilbertchen:
[...]

PR #244, opened 2017-10-21
Fix security weakness in storage key derivation

gilbertchen:
As suggested in #137, the key to encrypt the `config` file should be generated using a random salt and a configurable number of iterations.  This PR implements the required changes.

The new format for the `config` file is "duplicacy\001" + salt + #iterations + encrypted content, while the old format is "duplicacy\000" + encrypted content.

A `config` file in the old format can still be opened.  The `password` command can be used to convert to the new format (in addition to changing the password).

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -13,10 +13,11 @@ USAGE:
 
 OPTIONS:
    -encrypt, -e                    encrypt the storage with a password
-   -chunk-size, -c 4M              the average size of chunks
-   -max-chunk-size, -max 16M       the maximum size of chunks (defaults to chunk-size * 4)
-   -min-chunk-size, -min 1M        the minimum size of chunks (defaults to chunk-size / 4)
-   -pref-dir <preference directory path>    Specify alternate location for .duplicacy preferences directory
+   -chunk-size, -c <size>          the average size of chunks (default is 4M)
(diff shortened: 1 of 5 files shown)


robbat2:
How do you make it write out and old-format config file for compatibility?

gilbertchen:
You can always run the old version binary to create the old-format config file.

uup4:
Adding this here to link to #137. I made a further enhancement request comment on issue #137 for an option to allow the iteration count to be left out of the config file and be treated as user input. This is similar idea to PIM in VeraCrypt. So, a format something like "duplicacy\001" + salt + 0 + encrypted content, could mean that the actual count will be requested from the user with the password. A iteration count value > 0 in the config file would mean the user iteration count is stored in the config file.

uup4:
FYI..I took a quick glance at the code. In the below snippet, did you mean to use iterations and not iteartions as the name of int.

func GenerateKeyFromPassword(password string, salt []byte, iteartions int) []byte {

	return pbkdf2.Key([]byte(password), salt, iteartions, 32, sha256.New)

}

gilbertchen:
Yes, that is a typo.  Thanks for catching it!

uup4:
Awesome! Thank you.

________________________________

PR #254, opened 2017-10-30
 allow skip setting uid/gid on restored files

lowne:
> If the `-ignore-owner` option is specified, the restore procedure will not attempt to restore the original user/group id ownership on restored files (all restored files will be owned by the current user); this can be useful when restoring to a new or different machine.

Guidance needed about https://github.com/gilbertchen/duplicacy/compare/master...lowne:restore-ignore-uid?expand=1#diff-8542592c8489202c10c8a767962adc17L881 - unsure why that line is there.


gilbertchen:
You're right.  Changing metadata should also change the modified time, so if the modified time is unchanged there is no need to set the metadata.

I'll remove that line in a follow-up commit.

Issue #222, opened 2017-10-03
Feedback wanted:  new chunk directory structure


gilbertchen:
Currently the `chucks` directory is organized differently on different storages:
* For local/samba/sftp storages, chunks are stored hierarchically starting at level 2; adaptive nesting is enabled so if a subdirectory has too much files new subdirectories at a higher level will be created
* Dropbox storage also implements adapting nesting but starts at level 1
* All other storages use a flat structure -- chunks are stored directly under the `chunks` directory

The main problem with different structures is the incompatibility with rsync/rclone.  For example, it is not easy to sync a cloud storage to a local storage using third-party tools without 'flatting' the `chunks` directory, and vice versa.  The only choice is to run the copy command, which isn't as feature-rich/efficient as rsync/rclone (the only advantage of the copy command is the ability to cherry-pick a few backups instead of all backups to copy over).

There are other minor problems:
* sftp with a minimum level of 2 is too slow (65536 subdirectories to list)
* it is hard to browse or locate chunks with a flat structure on a web page 

Here is my proposal to overhaul the chunk hierarchy:

* A new file that can be manually created and edited, called `nesting`, defines the nesting levels
* This `nesting` file sits next to the `config` file on the storage, but is always unencrypted
* This file is a json file that has two keys: `read-levels` and `write-level`
* The value of `read-levels` is an array of integers representing which levels the chunks should be read from.  For example, `"read-levels": [0, 1, 2]` means chunks must be searched at level 0, level 1, and level 2, in that order
* The value of `write-level` is an integer representing the level that chunks should be written to.  For example `"write-level": 1` means new chunks will be written to level 1.
* In the new version, all storage backends will respect the levels defined in this file.  If this file is not present or if it is an old version of Duplicacy all storages will use the old hierarchy.

I think this solution not only ensures rsync/rclone interoperability, but also gives the ultimate freedom to users on how to organize the chunk directory.  It is better that adaptive nesting for two reasons:
* Most cloud storages can't support adaptive nesting because they don't have the notion of directories so they can't check if the directory for the next level exists, but now users can define an arbitrary level for any cloud storage
* This is more efficient because the number of lookups is equal to the number of read levels, while in adaptive nesting you need extra calls to check if directories exist.

dgcom:
Sorry for possibly stupid question, but:

> All other storages use a flat structure -- chunks are stored directly under the chunks **directory**

and

[...]

PR #259, opened 2017-11-07
Implement new chunk directory structure

gilbertchen:
This PR implements the new chunk directory structure proposed in #222:

* For storages initialized by earlier versions, there are no changes.
* New storages initialized now will have a key `fixed-nesting` set to `true` in the `config` file.
* This key forces all chunks, regardless of the storage type, to be saved with a nesting level of 1, so a chunk with an id `abcdef...` will be saved as `chunks\ab\cdef...`.
* The nesting level can be adjusted by adding a json file named `nesting` to the storage next to the `config` file, which should include two keys: `read-levels` and `write-level`
* The value of read-levels is an array of integers representing which levels the chunks should be read from. For example, `"read-levels": [0, 1, 2]` means chunks must be searched at level 0, level 1, and level 2, in that order
* The value of write-level is an integer representing the level that chunks should be written to. For example `"write-level": 1` means new chunks will be written to level 1.

### DIFF
--- a/src/duplicacy_gcdstorage.go
+++ b/src/duplicacy_gcdstorage.go
@@ -25,17 +25,18 @@ import (
 )
 
 type GCDStorage struct {
-	RateLimitedStorage
+	StorageBase
 
 	service     *drive.Service
 	idCache     map[string]string
-	idCacheLock *sync.Mutex
+	idCacheLock sync.Mutex
 	backoffs    []int // desired backoff time in seconds for each thread
 	attempts    []int // number of failed attempts since last success for each thread
 
-	isConnected     bool
-	numberOfThreads int
-	TestMode        bool
+	createDirectoryLock sync.Mutex
+	isConnected         bool
+	numberOfThreads     int
+	TestMode            bool
 }
 
 type GCDConfig struct {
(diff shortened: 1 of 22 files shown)


src/duplicacy_gcdstorage.go
```diff
@@ -114,6 +119,9 @@ func (storage *GCDStorage) shouldRetry(threadIndex int, err error) (bool, error)
 	return true, nil
 }
 
+// convertFilePath converts the path for a fossil in the form of 'chunks/id.fsl' to 'fossils/id'.  This is because
+// ACD doesn't support file renaming. Instead, it only allows one file to be moved from one directory to another.
```
TheBestPessimist: Here it should say GCD

src/duplicacy_gcdstorage.go
```diff
@@ -328,8 +358,9 @@ func CreateGCDStorage(tokenFile string, storagePath string, threads int) (storag
 
 	storage.isConnected = true
 
+	storage.DerivedStorage = storage
+	storage.SetDefaultNestingLevels([]int{0}, 0)
```
TheBestPessimist: I don't understand most of what you did here so this question may be stupid, but why here the default nesting levels is 0 and 0, instead of the "default" 1 and 1? Note that i also saw this nesting for other storages unrelated to google: s3cstorage, or azurestorage.

gilbertchen: This is the backward-compatible default, which could be either 2,3 and 2, or 0 and 0, depending on the storage type, so it has to be set in the constructor.  The new "default" 1 and 1 is the same for all storage, so it can be set in SetNestingLevels.

TheBestPessimist: In that case, it is too much to make a note of this somewhere? Let's say maybe in the class (is it a class?) StorageBase?

gilbertchen:
Thank you for reviewing it.  I'll make suggested changes in a separate checkin.

TheBestPessimist:
Sorry for not doing more. This is more than a little over my skills :(

lowne:
@gilbertchen is there any way to add the `FixedNesting` flag on existing (encrypted, and compliant) storages? 😭

gilbertchen:
@lowne there is no straightforward way, but I think you can run `add` to create a new storage and then run `copy` to copy everything over.  If it is an SFTP server you can run these two commands locally.

lowne:
@gilbertchen that wouldn't work for me for a number of reasons. However I solved via #264 🎉

Issue #260, opened 2017-11-08
SSH agent registered, but still asking for private key file


andrej-zirko:
Hello,
I'm using SSH agent with my password protected private key. Although my SSH agent is running fine duplicacy is asking me to enter private key file

> Enter the path of the private key file:

When I enter empty string duplicacy runs just fine. Problem is that this is very confusing for the user. I had to look at the duplicacy code to figure out that I should enter empty value for the prompt.

Is this a bug or feature? Logically I would put a condition 
`if len(signers) > 0` and return after agentSock determination. This would skp the private key file prompt. But maybe I miss something. 

https://github.com/gilbertchen/duplicacy/blob/7e1fb6130a3cc7994e80400843fde63ca288c1e1/src/duplicacy_storage.go#L232-L250

gilbertchen:
Your change will not work in the case where the ssh-agent is set up but no key has been added.  `len(signers)` will not be zero in this case, but since no key is provided the authentication will fail and there will be no way for the user to enter the key file.

So I think we just need to state in the user guide that to enable login via ssh-agent, an empty private key file should be entered.

ddribin:
I just hit this on version 2.5.1 trying to backup a Linux machine over `sftp`. So I did this:

    touch ~/.duplicacy/ssh_key_file
    export DUPLICACY_SSH_KEY_FILE=$HOME/.duplicacy/ssh_key_file

And while the backup does work, I get this warning:

    Failed to parse the private key file /home/dave/.duplicacy/ssh_key_file: ssh: no key found

Is there any way to get rid of that warning?

PR #261, opened 2017-11-08
don't ask for keyFile if SSH agent provided signers

andrej-zirko:


gilbertchen:
Per my comment in #260 I'm closing this PR without merging it.

PR #262, opened 2017-11-09
added trailing "/" to sftp storage dir, if omitted

andrej-zirko:


gilbertchen:
Sorry I can't merge this PR as it will break backward compatibility.

Issue #280, opened 2017-11-29
enhancement : create destination folder if it doesn't exist


michaelcinquin:
Hi

I need to keep my destination folder as small as possible, because I need to run `duplicacy check` on them, because during my tests I found a successful backup to Gdrive that yields an impossible restore due to a missing chunk (??).

But duplicacy check takes ages, because it lists every chunk from every snapshot, even though I specified the -id option.

So, one solution would be to keep everything separate on the storage (one folder backup-up on source = one folder on storage). But for this, duplicacy needs to create the destination folder if it doesn't exist.  Of course, de-duplication is less effective then.

So I modified duplicacy_gcdstorage.go, turned line 336
storagePathID, err := storage.getIDFromPath(0, storagePath, false)
into
storagePathID, err := storage.getIDFromPath(0, storagePath, true)

I think it's sensible to do createDirectories if they don't exist on cloud storage, don't you?

gilbertchen:
It should be okay to create the destination folder on the storage if it doesn't exist.  At least it won't break anything.  If you can work on a pull request I'll merge it.

On the other hand, we'll still require users to manually create the destination folder before every storage backend implements this.

PR #283, opened 2017-12-01
create the destination folder on gcd storage if it doesn't exist

michaelcinquin:
as discussed on https://github.com/gilbertchen/duplicacy/issues/280#issuecomment-348400492

allow the creation of the destination folder on gcd storage if it doesn't exist


gilbertchen:
Thank you for the fix!

PR #287, opened 2017-12-04
Add a .gitignore file

TheBestPessimist:
Since the project doesn't have one, maybe a .gitignore file could be useful.
Right now, this file handles things generated by intellijIdea and also by some other go tools.


TheBestPessimist:
Wrong branch

PR #288, opened 2017-12-04
Add a .gitignore file

TheBestPessimist:
Since the project doesn't have one, maybe a .gitignore file could be useful.
Right now, this file handles things generated by intellijIdea and also by some other go tools.


TheBestPessimist:
If the ignore file is too eager and ignores too much we could just comment/remove some lines from it at a later time.

In this version however, it only excludes things from intellij Idea, which are generally considered  as "should not put under VCS", and therefore i think it will not break anything.

gilbertchen:
I still think it is not a good idea to check in a .gitignore file, since it can be different for everyone and that would create extra burden when committing changes.  I'm closing this PR for now but if you can persuade me I'll reopen it.

PR #328, opened 2018-01-13
gitignore vendor

pdf:
The vendor dir should either be checked into the repo, or ignored by git.  Since the former was not true, the latter is applied here.


pdf:
The explicit purpose of `.gitgnore` is to ignore project-specific paths.  Users can add global ignores via their git config if they want.

pdf:
Also, I can't globally ignore `vendor`, and I can't ignore `.gitignore`, which means that one way or another there are files floating around in the project dir that are not checked in, which makes working with git significantly more error-prone and noisy.

PR #329, opened 2018-01-14
Reduce memory consumption for prune operation

### LINKED ISSUE #146: Prune excessive memory consumption
I have a 1GB RAM machine that generates one revision per day, with a total local storage volume of 12GB, and a retention policy of `7:30`, and `0:120`.  Attempting to perform a prune operation with ~900MB free RAM results in the process being OOM killed after it consumes all available memory.

pdf:
This is a first pass at fixing prune memory consumption.  I'm not as familiar with the codebase as I would like, but I think this is in a state that's okay for review.  I couldn't get any of the tests to work locally though, so I'm not sure how to test further except to run against valid data (which I've done for the standard prune, but not for exhaustive/exclusive).

Implements non-caching variants of `GetSnapshotChunks`/`LoadChunks`. All existing call sites are migrated to use `CacheSnapshotChunks` and `CacheChunks`, which are the renamed (previously existing) caching implementations.  This allows individual callers to be updated to use the non-caching version where that is deemed appropriate. This is a breaking change for any third-party tooling, as the return signature for `LoadChunks` has changed, and `GetSnapshotChunks` no longer populates  `snapsnot.ChunkHashes`.

Updates `PruneSnapshots` to use the non-caching chunk fetching functions.  For non-exhaustive prune, considers only target chunks instead of mapping all chunks in repository to memory.

Fixes #146

PS - I should also note that there are a lot of unhandled errors in the existing code-base, plus variable shadowing, etc.  In the code that I've touched here, I've made sure to at least log errors where I wasn't 100% certain that failing was correct, and fixed lint where found.  I'd highly recommend incorporating some linters into your workflow to catch these sorts of things.

### DIFF
--- a/src/duplicacy_snapshotmanager.go
+++ b/src/duplicacy_snapshotmanager.go
@@ -387,7 +387,7 @@ func (manager *SnapshotManager) CleanSnapshotCache(latestSnapshot *Snapshot, all
 	chunks := make(map[string]bool)
 
 	if latestSnapshot != nil {
-		for _, chunkID := range manager.GetSnapshotChunks(latestSnapshot) {
+		for _, chunkID := range manager.GetSnapshotChunks(latestSnapshot, false) {
 			chunks[chunkID] = true
 		}
 	}
@@ -597,8 +597,9 @@ func (manager *SnapshotManager) ListAllFiles(storage Storage, top string) (allFi
 	return allFiles, allSizes
 }
 
-// GetSnapshotChunks returns all chunks referenced by a given snapshot.
-func (manager *SnapshotManager) GetSnapshotChunks(snapshot *Snapshot) (chunks []string) {
+// GetSnapshotChunks returns all chunks referenced by a given snapshot. If
+// keepChunkHashes is true, snapshot.ChunkHashes will be populated.
+func (manager *SnapshotManager) GetSnapshotChunks(snapshot *Snapshot, keepChunkHashes bool) (chunks []string) {
 
 	for _, chunkHash := range snapshot.FileSequence {
 		chunks = append(chunks, manager.config.GetChunkIDFromHash(chunkHash))
@@ -625,6 +626,10 @@ func (manager *SnapshotManager) GetSnapshotChunks(snapshot *Snapshot) (chunks []
 
 	for _, chunkHash := range snapshot.ChunkHashes {
 		chunks = append(chunks, manager.config.GetChunkIDFromHash(chunkHash))
+	}
+
+	if !keepChunkHashes {
+		snapshot.ClearChunks()
 	}
 
 	return chunks
@@ -715,7 +720,7 @@ func (manager *SnapshotManager) ListSnapshots(snapshotID string, revisionsToList
 			}
 
 			if showChunks {
-				for _, chunkID := range manager.GetSnapshotChunks(snapshot) {
+				for _, chunkID := range manager.GetSnapshotChunks(snapshot, false) {
 					LOG_INFO("SNAPSHOT_CHUNKS", "chunk: %s", chunkID)
 				}
 			}
@@ -805,7 +810,7 @@ func (manager *SnapshotManager) CheckSnapshots(snapshotID string, revisionsToChe
 			}
 
 			chunks := make(map[string]bool)
-			for _, chunkID := range manager.GetSnapshotChunks(snapshot) {
+			for _, chunkID := range manager.GetSnapshotChunks(snapshot, false) {
 				chunks[chunkID] = true
 			}
 
@@ -895,7 +900,7 @@ func (manager *SnapshotManager) ShowStatistics(snapshotMap map[string][]*Snapsho
 		for _, snapshot := range snapshotList {
 
 			chunks := make(map[string]bool)
-			for _, chunkID := range manager.GetSnapshotChunks(snapshot) {
+			for _, chunkID := range manager.GetSnapshotChunks(snapshot, false) {
 				chunks[chunkID] = true
 				snapshotChunks[chunkID] = true
 			}
@@ -948,7 +953,7 @@ func (manager *SnapshotManager) ShowStatisticsTabular(snapshotMap map[string][]*
 		earliestSeenChunks := make(map[string]int)
 
 		for _, snapshot := range snapshotList {
-			for _, chunkID := range manager.GetSnapshotChunks(snapshot) {
+			for _, chunkID := range manager.GetSnapshotChunks(snapshot, true) {
 				if earliestSeenChunks[chunkID] == 0 {
 					earliestSeenChunks[chunkID] = math.MaxInt32
 				}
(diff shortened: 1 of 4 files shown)


pdf:
I apologise for the somewhat hard to follow git diff, it comes mostly from refactoring out the actual prune operations into `pruneSnapshots` and `pruneSnapshotsExclusive`, rather than keeping them inside the huge `if`/`else` statement.

pdf:
I've done some further testing and profiling. Looks like mostly by virtue of using the non-caching chunk methods, exhaustive prune memory usage tops out at around 80MB, and standard prune at around 65MB for my dataset.  The latter is down from somewhere well over 1GB on master.

If we determine that this code is correct, it's probably worth examining some of the other usages to determine where chunk caching is actually beneficial, as there are likely other places where significant savings could me made, just by switching to the non-caching methods.

src/duplicacy_snapshotmanager.go
```diff
@@ -597,8 +598,8 @@ func (manager *SnapshotManager) ListAllFiles(storage Storage, top string) (allFi
 	return allFiles, allSizes
 }
 
-// GetSnapshotChunks returns all chunks referenced by a given snapshot.
-func (manager *SnapshotManager) GetSnapshotChunks(snapshot *Snapshot) (chunks []string) {
+// CacheSnapshotChunks returns all chunks referenced by a given snapshot.
+func (manager *SnapshotManager) CacheSnapshotChunks(snapshot *Snapshot) (chunks []string) {
```
gilbertchen: I think instead of keeping two versions of GetSnapshotChunks and LoadChunks, it is possible to add an argument to the original GetSnapshotChunks that indicates if `snapshot.ChunkHashes` needs to be cleared at the end of the call, like:

```
func (manager *SnapshotManager) GetSnapshotChunks(snapshot *Snapshot, clearChunkHashes bool) (chunks []string) {
```

The only place where `clearChunkHashes` should be false is inside `ShowStatisticsTabular` where `GetSnapshotChunks` is called twice on all snapshots.  In all other places I believe it is safe to set `clearChunkHashes` to true.

src/duplicacy_snapshotmanager.go
```diff
@@ -1585,15 +1616,24 @@ func (manager *SnapshotManager) PruneSnapshots(selfID string, snapshotID string,
...
+	if err != nil {
+		LOG_ERROR("LOG_FILE", "Could not open log file %s: %v", logFileName, err)
+	}
 
 	defer func() {
 		if logFile != nil {
-			logFile.Close()
+			cerr := logFile.Close()
+			if cerr != nil {
+				LOG_ERROR("LOG_FILE", "Could not close log file %s: %v", logFileName, cerr)
```
gilbertchen: This should be LOG_WARN, because this defer func can be called when a panic occurs and LOG_ERROR would cause a new panic.

src/duplicacy_snapshotmanager.go
```diff
@@ -1910,18 +1961,101 @@ func (manager *SnapshotManager) PruneSnapshots(selfID string, snapshotID string,
...
+	if toBeDeleted == 0 && !exhaustive {
 		LOG_INFO("SNAPSHOT_NONE", "No snapshot to delete")
 		return false
 	}
 
-	chunkRegex := regexp.MustCompile(`^[0-9a-f]+$`)
+	collection := CreateFossilCollection(allSnapshots)
 
-	referencedChunks := make(map[string]bool)
+	var success bool
```
gilbertchen: Thank you for splitting this into separate functions!  I have been too lazy and too afraid of breaking something.  Don't worry about the git diff results.  I'm sure a manual diff of the relevant parts should be good.

src/duplicacy_snapshotmanager.go
pdf: I considered both approaches (parameter vs separate methods).
- The logic for `GetSnapshotChunks` using a parameter is not too complicated, I'd probably name the parameter something like `cacheChunks` and reverse the logic though, since the zero value for bool is `false`, and the default behaviour sounds like it should be not to cache the chunks in memory.
- The problem for me is with usage of `LoadChunks`, since this is used directly in a few places.  If this is retained in its original state, storing data on the snapshot object, it becomes the responsibility of every calling site to `nil` that field if they don't want to cache the chunks, which is just asking for unexpected memory usage.  `LoadChunks` also doesn't really lend itself well to operating on a parameter for caching, since it needs to return the chunks for the non-caching version, but also store them on the snapshot when caching.

If you would prefer not to retain two implementations, one option would be to replace the caching variant of `LoadChunks` with the implementation in this PR, which will return the chunks without storing them, then store them on the snapshot inside `GetSnapshotChunks`, based on a parameter.  The reason I didn't consider this route is because I wasn't sure how `LoadChunks` was used in places like `DownloadSnapshotSequence`, and I didn't have time to work back up the call chain(s) for these other methods.

src/duplicacy_snapshotmanager.go
pdf: Ack

src/duplicacy_snapshotmanager.go
pdf: No problem - it made particular sense to do it now, as the two paths now also differ in how they walk the snapshots.

pdf:
If you'd prefer to run with this yourself, you're quite welcome, maintainer edits are enabled on the PR.  However, I'm happy to keep working on this based on your feedback if you'd like me to, as time allows.

src/duplicacy_snapshotmanager.go
gilbertchen: What I thought was to keep `LoadChunks` as it is.  Other than `GetSnapshotChunks`, it is only called by `DownloadSnapshotSequence`, which in turn is called by `DownloadSnapshotContents`.  It should be okay to keep `snapshot.ChunkHashes` in ``DownloadSnapshotContents` I think.

Feel free to reverse the logic for `clearChunkHashes` and give it a different name; just avoid using `cache` in its name.  This would create some confusion with the snapshot cache.

gilbertchen:
Ok, please keep working on this.  I really appreciate your help!

pdf:
Updated based on feedback, ready for review again.

pdf:
I need to do some more profiling on this - I'm not certain that this results in the same level of memory reduction as the previous implementation.

pdf:
This code is not sufficient - there are locations that assume stored hashes that are not handled here, I'll take another pass.

pdf:
Okay, I _think_ it's just `*BackupManager.Backup` that requires `keepChunkHashes` enabled, but I'm out of time trying to back-track all the references to `*Snapshot.ChunkHashes`.  It might be safer just to enable keeping chunks everywhere initially, until someone who understands all the code-paths better than me can critically review and disable caching for each call to `GetSnapshotChunks`?

In the mean time, I've enabled keeping for chunks only for `Backup` and `ShowStatisticsTabular`.

jt70471:
I'm all for any improvements that can be made obviously, but pruning is such an important function, I hope this is heavily tested before it's committed to master... :)

Has any testing been done on two diff repositories to ensure that the results of prune operations using the current master code and with these modifications results in the exact same results?

Thanks

gilbertchen:
@pdf Thank you for your contribution!  This is a good piece of work and may have fixed a previously unknown bug!

@jt70471 I tested on my own repository and made sure they pruned the same sets of chunks.

kairisku:
It seems the emptying of `ChunkHashes` causes backup statistics to show incorrect data (0 total):

    File chunks: 0 total, 19,552M bytes; 102 new, 82,075K bytes, 49,195K bytes uploaded

gilbertchen:
The 0 file chunk bug should be fixed by https://github.com/gilbertchen/duplicacy/commit/5d2242d39d7b709b7a5d1e286c326b0f2d6143ca.

Issue #322, opened 2018-01-07
Fossilization incurs additional costs on Wasabi


markfeit:
(This is mostly a placeholder; I may work on this myself as an excuse to learn Go.  A more-detailed explanation has been added to the [Storage Backends](https://github.com/gilbertchen/duplicacy/wiki/Storage-Backends) page on the wiki.)

S3 lacks support for renaming objects and instead accomplishes it with a copy and delete.  When used with Wasabi, this incurs additional costs because Wasabi's billing model charges for a minimum of 90 days of storage for any created object, and copies create new objects with new 90-day clocks.

Wasabi's API has a rename operation that can be performed at no cost and leaves the object in its original state with respect to billing.  Separate code for Wasabi that incorporates this should be developed.

gilbertchen:
Thanks for the information.  I wasn't aware of this.  I think you just need to add a new parameter to the `S3Storage` struct to indicate if there is a rename api.  And then add a new storage url prefix `wasabi://` for which the parameter will be `true` but `false` for the regular `s3://` storage.

markfeit:
There are a couple of reasons why I don't think it's going to be that simple, but feel free to correct me if I'm missing something.

First, AWS doesn't have rename, so there's no standard API for it.  Wasabi borrows WebDAV's `MOVE` verb, but there's no guarantee that's going to be the same on other services that have an API of their own.

Second, there's no support for it in [aws-sdk-go](https://github.com/aws/aws-sdk-go).  As much as I'd like to use as much of the existing S3 code as I can, it doesn't look like it would be easy to make the SDK do it without a fork of the whole thing.  Amazon would probably not be inclined to take a pull request that helps out a competitor.

The subset of the SDK used by Duplicacy is small enough that it might be worth doing a standalone module as was done for B2 and either make it Wasabi-only or give it a switch as you suggested and use it for everything S3-ish.

gilbertchen:
Then a Wasabi-only storage backend would make sense. 

But which S3 client library are you going to use?  Or are you thinking of calling the API directly just like how B2Storage is implemented?  I know Wasabi is using the [minio server](https://github.com/minio/minio) internally, and there is a minio client library too (which however doesn't seem to support the MOVE/RENAME method).

markfeit:
[...]

PR #331, opened 2018-01-17
Added semi-dedicated Wasabi storage module.  #322

markfeit:
Adds a storage module for Wasabi that uses the existing S3 code for everything except renaming objects.  See #322.

### DIFF
--- a/src/duplicacy_storage.go
+++ b/src/duplicacy_storage.go
@@ -461,6 +461,38 @@ func CreateStorage(preference Preference, resetPassword bool, threads int) (stor
 		SavePassword(preference, "s3_secret", secretKey)
 
 		return storage
+
+	} else if matched[1] == "wasabi" {
+
+		region := matched[2]
+		endpoint := matched[3]
+		bucket := matched[5]
+
+		key := GetPassword(preference, "wasabi_key",
+			"Enter Wasabi key:", true, resetPassword)
+		secret := GetPassword(preference, "wasabi_secret",
+			"Enter Wasabi secret:", true, resetPassword)
+
+		storageDir := ""
+		if strings.Contains(bucket, "/") {
+			firstSlash := strings.Index(bucket, "/")
+			storageDir = bucket[firstSlash+1:]
+			bucket = bucket[:firstSlash]
+		}
+
+		storage, err := CreateWasabiStorage(region, endpoint,
+			bucket, storageDir, key, secret, threads)
+
+		if err != nil {
+			LOG_ERROR("STORAGE_CREATE", "Failed to load the S3 storage at %s: %v", storageURL, err)
+			return nil
+		}
+
+		SavePassword(preference, "key", key)
+		SavePassword(preference, "secret", secret)
+
+		return storage
+
 	} else if matched[1] == "dropbox" {
 		storageDir := matched[3] + matched[5]
 		token := GetPassword(preference, "dropbox_token", "Enter Dropbox access token:", true, resetPassword)
--- a/src/duplicacy_storage.go
+++ b/src/duplicacy_storage.go
@@ -468,6 +468,10 @@ func CreateStorage(preference Preference, resetPassword bool, threads int) (stor
 		endpoint := matched[3]
 		bucket := matched[5]
 
+		if region != "" {
+			region = region[:len(region)-1]
+		}
+
 		key := GetPassword(preference, "wasabi_key",
 			"Enter Wasabi key:", true, resetPassword)
 		secret := GetPassword(preference, "wasabi_secret",
--- a/src/duplicacy_storage.go
+++ b/src/duplicacy_storage.go
@@ -484,12 +484,12 @@ func CreateStorage(preference Preference, resetPassword bool, threads int) (stor
 			bucket, storageDir, key, secret, threads)
 
 		if err != nil {
-			LOG_ERROR("STORAGE_CREATE", "Failed to load the S3 storage at %s: %v", storageURL, err)
+			LOG_ERROR("STORAGE_CREATE", "Failed to load the Wasabi storage at %s: %v", storageURL, err)
 			return nil
 		}
 
-		SavePassword(preference, "key", key)
-		SavePassword(preference, "secret", secret)
+		SavePassword(preference, "wasabi_key", key)
+		SavePassword(preference, "wasabi_secret", secret)
 
 		return storage
 
--- a/src/duplicacy_storage_test.go
+++ b/src/duplicacy_storage_test.go
@@ -78,10 +78,12 @@ func loadStorage(localStoragePath string, threads int) (Storage, error) {
 		storage, err := CreateSFTPStorageWithPassword(config["server"], port, config["username"], config["directory"], 2, config["password"], threads)
 		storage.SetDefaultNestingLevels([]int{2, 3}, 2)
 		return storage, err
-	} else if testStorageName == "s3" || testStorageName == "wasabi" {
+	} else if testStorageName == "s3" {
 		storage, err := CreateS3Storage(config["region"], config["endpoint"], config["bucket"], config["directory"], config["access_key"], config["secret_key"], threads, true, false)
 		storage.SetDefaultNestingLevels([]int{2, 3}, 2)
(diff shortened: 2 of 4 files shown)


src/duplicacy_wasabistorage.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+func (storage *WasabiStorage) DeleteFile(
+	threadIndex int, filePath string,
+) (err error) {
+	return storage.s3.DeleteFile(threadIndex, filePath)
+
+}
+
+// This is a lightweight implementation of a call to Wasabi for a
```
gilbertchen: This is clever!  much simpler than I thought.

src/duplicacy_wasabistorage.go
```diff
@@ -0,0 +1,189 @@
...
+	if error != nil {
+		return error
+	}
+	request.Header.Add("Authorization", authorization)
+	request.Header.Add("Date", timestamp)
+	request.Header.Add("Destination", new_name)
+	request.Header.Add("Host", storage.endpoint)
+	request.Header.Add("Overwrite", "true")
+
+	client := &http.Client{}
```
gilbertchen: I think `client` needs to be a field of `WasabiStorage` so it can reuse https connections when possible.  Otherwise there would be too much overhead to reestablish the https connection every time a new request is sent.

src/duplicacy_wasabistorage.go
```diff
@@ -0,0 +1,189 @@
...
+
+	response, error := client.Do(request)
+	if error != nil {
+		return error
+	}
+
```
gilbertchen: `response` should be closed if the connection is to be reused.

src/duplicacy_wasabistorage.go
```diff
@@ -0,0 +1,189 @@
...
+	if response.StatusCode != 200 {
```
gilbertchen: I would like to see some retry logic in the function in the future; it doesn't need to be in this PR -- adding a comment here as a reminder should be good.

markfeit: Retry logic would be better higher up in the program with the storage modules being given the number of attempts so far and returning pass/fail and, in the latter case, how long to wait before the next retry or a signal to give up.  (MHO, YMMV)

markfeit:
Is anything still holding this up, or are we good?  (FWIW, I've replaced S3 with this patch for my own backups and it seems fine.)

gilbertchen:
I think it is good but I haven't run any tests myself.  I'm working on 2.1.0 release and just to be on the safe side I'll merge this after the 2.1.0 release is out.

src/duplicacy_wasabistorage.go
```diff
@@ -0,0 +1,190 @@
...
+// general-purpose and reusable.
+func (storage *WasabiStorage) MoveFile(
+	threadIndex int, from string, to string,
+) (err error) {
+
+	// The from path includes the bucket
+	from_path := fmt.Sprintf("/%s/%s/%s", storage.bucket, storage.storageDir, from)
+
+	// The to path does not.
+	// to_path := fmt.Sprintf("%s/%s", storage.storageDir, from)
```
gilbertchen: Please delete this line.

src/duplicacy_wasabistorage.go
```diff
@@ -0,0 +1,190 @@
...
+
+	// We get the region with an @ on the end, so don't add another.
+	object := fmt.Sprintf("https://%s%s%s",
```
gilbertchen: I think there should be a `.` between the region and the host name, like `us-east1.wasabysys.com`.

src/duplicacy_storage_test.go
```diff
@@ -78,10 +78,12 @@ func loadStorage(localStoragePath string, threads int) (Storage, error) {
 		storage, err := CreateSFTPStorageWithPassword(config["server"], port, config["username"], config["directory"], 2, config["password"], threads)
 		storage.SetDefaultNestingLevels([]int{2, 3}, 2)
 		return storage, err
-	} else if testStorageName == "s3" || testStorageName == "wasabi" {
+	} else if testStorageName == "s3" {
 		storage, err := CreateS3Storage(config["region"], config["endpoint"], config["bucket"], config["directory"], config["access_key"], config["secret_key"], threads, true, false)
 		storage.SetDefaultNestingLevels([]int{2, 3}, 2)
```
gilbertchen: Missing a return statement here.

src/duplicacy_storage_test.go
```diff
@@ -78,10 +78,12 @@ func loadStorage(localStoragePath string, threads int) (Storage, error) {
...
-		return storage, err
+	} else if testStorageName == "wasabi" {
+		storage, err := CreateWasabiStorage(config["region"], config["endpoint"], config["bucket"], config["directory"], config["access_key"], config["secret_key"], threads)
+		storage.SetDefaultNestingLevels([]int{2, 3}, 2)
```
gilbertchen: Missing a return statement here.

src/duplicacy_wasabistorage.go
markfeit: Per the comment at line 106, the region passed in by the application has an `@` on the end, so the formatted URL ends up `us-east-1@s3.wasabisys.com`, which is acceptable.

Note that the behavior between the application and the test code is inconsistent; the latter passes the region name _without_ the `@` where the regular expression in the former collects it.  I've adjusted the storage module to add it if it's not there, but making it consistent globally would be worth writing an issue.

markfeit:
Following an additional cleanup to the test code and storage module to correct for an inconsistency (see comment above):

```
$ make test
cd gopath/src/github.com/gilbertchen/duplicacy \
&& GOPATH=/local/export/home/mfeit/work/duplicacy-dev/gopath go test -run TestStorage ./src -storage wasabi
ok      github.com/gilbertchen/duplicacy/src    2.361s
cd gopath/src/github.com/gilbertchen/duplicacy \
&& GOPATH=/local/export/home/mfeit/work/duplicacy-dev/gopath go test -run TestBackupManager ./src -storage wasabi
ok      github.com/gilbertchen/duplicacy/src    10.126s
```

src/duplicacy_storage.go
```diff
@@ -461,6 +461,38 @@ func CreateStorage(preference Preference, resetPassword bool, threads int) (stor
 		SavePassword(preference, "s3_secret", secretKey)
 
 		return storage
+
+	} else if matched[1] == "wasabi" {
+
+		region := matched[2]
+		endpoint := matched[3]
+		bucket := matched[5]
+
```
gilbertchen: The s3 storage strips off the trailing @ from the region.  We should do the same here.

src/duplicacy_wasabistorage.go
```diff
@@ -0,0 +1,197 @@
...
+// descriptions.
+
+func CreateWasabiStorage(
+	regionName string, endpoint string,
+	bucketName string, storageDir string,
+	accessKey string, secretKey string,
+	threads int,
+) (storage *WasabiStorage, err error) {
+
+	// The application code that parses the URL from the
```
gilbertchen: I think the region name shouldn't contain a `@`.  My main concern is that passing such a region name to `s3storage` may create some trouble.  Besides, it doesn't look good if you print the region name in log messages.

If you make the other change to take out the `@` from the regex matching result we can assume that the region name passed in will never contain a `@`.

src/duplicacy_storage.go
markfeit: Ah, I see what's going on: the root of the problem is the [regexp that picks the URL apart](https://github.com/gilbertchen/duplicacy/blob/e07226bd621dcf03b15193e60ecbcb9efd3e797c/src/duplicacy_storage.go#L264) collecting the `@` and anything afterward consuming it has to compensate.

I've adjusted the Wasabi code to make the correction so it's at least consistent with everything else.

Better would be to correct the problem at the root and not collect the `@` in the first place so it doesn't have to be band-aided later.  Better still would be to use `net/url` to parse the URL.  Both would require modifications outside the scope of this PR.

gilbertchen:
Thank you for this contribution!

Issue #384, opened 2018-03-18
Can Duplicacy CLI have a -comment qualifier?


jeffaco:
I'm looking at scripting `duplicacy` on my Mac system.

Some basics:
1. I have several different repositories that I wish to back up (about 3 of them).
2. I just don't entirely trust cloud storage providers, so I have each backup on two independent cloud storage providers.
3. I want _independent_ backups to run concurrently (here, _independent_ means that they are backing up completely different repositories).

I have the regular set of tools available to me via bash scripting (what I'm most comfortable with given my Linux/UNIX background) like ps, grep, awk, etc.

I have no easy way to determine, using my set of tools, what repository is being backed up by `duplicacy`, so I'm having a problem implementing my scripting. It's trivial if I want no concurrent `duplicacy` jobs, but jobs on two independent repositories are completely safe to run.

Could `duplicacy` have a `-comment` qualifier that can contain any text I want? I'm thinking something like `duplicacy -comment officeDB backup -threads 2`? I like the idea of -comment being at the top level (not just for backup) so I could avoid concurrent operations of all sorts for a repository operations (prune, etc).

Thoughts?

The idea here: Parse this and completely ignore it.

Is there another way to achieve what I'm trying to do?

gilbertchen:
Nice idea.  I think you can use the `-tag` option for the backup command if you don't use it already, but there isn't anything similar for other commands.

jeffaco:
If I use the `-tag` option, that refers to a particular revision. Can I use the same tag, over and over, such that it would be assigned to multiple revisions? If not, then I don't think this would work.

I haven't done much with the Go language yet, but I'm thinking of learning it. Are you open to PRs if/when I get to start learning Go?

gilbertchen:
The same tag can be applied to more than one revision, but unlike `-comment` it does have the side effect of being stored in the snapshot files.

Yes, I'll appreciate it if you can add this option.

jeffaco:
I have a commit that, I believe, implements this. Easier than I thought! However, as a 'go' newbie, I'm not sure how to build Duplicacy and, much to my surprise, there wasn't a peep about it on the README.md page.

I have a GO workspace directory tree (`go/bin`, `go/src`, `go/package`). Under `go/src`, I have the duplicacy repository. But if I cd to there, I get:

```
$ go build
can't load package: package duplicacy: no Go files in /Users/jeff/local/gowork/src/duplicacy
$ 
```

[...]

Issue #361, opened 2018-02-15
Failed to upload the chunk / connection timed out errors using Azure


Kyle2123:
While running the initial backup from MacOS and Linux computers running 2.0.10 to Azure storage, the process eventually fails with a message like this:

> Failed to upload the chunk f041b71f8ae1e436ea2e6ce647c0491426a3b146bd2538308cf186fcea5d3e48: Put https://[my storage account].blob.core.windows.net/[my storage container]/chunks/f0/41b71f8ae1e436ea2e6ce647c0491426a3b146bd2538308cf186fcea5d3e48: read tcp 192.168.1.145:59117->52.239.153.4:443: read: connection timed out
Incomplete snapshot saved to /mnt/USBHD/.duplicacy/incomplete

Whenever this happens I just restart the backup process again.  Eventually, it finishes.  The problem is that these are the initial backups so they are huge and instead of it just running for a day, it's taking forever because it fails after an hour or two and doesn't get started up again until I notice it.

I am not sure if this happens on Windows or not because I haven't tried from a Windows PC yet but I assume it does.  I am also not sure if this only happens when writing to Azure storage or if it affects other clients too.

gilbertchen:
How many threads are you using?

Kyle2123:
I haven't specified the threads option so it is whatever the default is.  The command I am using is "duplicacy backup -stats"

gilbertchen:
Can you change this line: https://github.com/gilbertchen/duplicacy/blob/e07226bd621dcf03b15193e60ecbcb9efd3e797c/src/duplicacy_azurestorage.go#L11
to:
```
    "github.com/gilbertchen/azure-sdk-for-go/storage"
```
and then rebuild from the source?

This will pick up the change I made to retry on timeout errors: https://github.com/gilbertchen/azure-sdk-for-go/commit/53194c24dc6a5c95f92045f313c6fb719a8c24f8

If you would rather like a working binary, please let me know.

Kyle2123:
Sorry for the delayed response - was out of town last week.  I don't have an easy way to do the builds myself, can you help me get a binary to test this with (for Mac)?

jeffaco:
Hey, I'm having this identical problem, see [here](https://duplicacy.com/issue?id=5654313976201216) for details. I, too, am running on Mac OS/X.

Any problem if I test this fix as well?

gilbertchen:
@jeffaco please test the fix and let me know if it works.

[...]

PR #391, opened 2018-03-28
Add -comment capability to allow for duplicacy jobs to be identified with 'ps'

jeffaco:
@gilbertchen 

Two commits here:

1. Minor change to README.md to reference how to build `Duplicacy`.
2. Added `-comment` qualifier as per issue #384 

Given that you have changes I need not currently committed (identified in #361), a new build of duplicacy for the Mac with this change would be awesome, thanks in advance.


gilbertchen:
Thank you for adding this nice option!

PR #392, opened 2018-04-01
Add nobackup-file preference.

fhriley:
Directories containing a file with this name will not be backed up. I find it easier to drop a .nobackup file in directories I don't want backed up instead of maintaining a file of exclusions. This is also useful for scripts that create data in the repository but don't want it to be backed up.


gilbertchen:
Thank  you for your contribution!

gilbertchen:

https://forum.duplicacy.com/t/help-with-filters-again/8927/10

Issue #302, opened 2017-12-11
Add separate config file to force certain paths


markfeit:
Following onto #216:

I'm testing Duplicacy as a replacement for CrashPlan and have it set up in the root of a system:
```
[root@host /]# ls /
bin   cvs  etc     home  lib64  local1      media  opt   root  sbin  sys  usr
boot  dev  export  lib   local0 lost+found  mnt    proc  run   srv   tmp  var
```

Ultimately, I'm going to want to back up all of `/local0`, `/local1` (multiple terabytes), with everything else considered expendable.  That's easy to set up as a filter:
```
+local0/*
+local1/*
-*
```

For now, I just want to work with a small slice of my home directory that changes frequently and contains some files that exercise some of the other filters I'm going to use.  The only way I can find to make it happen is by filtering like this, which is [functional but ugly](http://www.osnews.com/story/19266/WTFs_m):
```
+local/export/home/me/var/
+local/export/home/me/
+local/export/home/
+local/export/
+local/                                                                        
-*
```
(I could also weed out everything except that, but that's uglier.)

I'd like to propose that an optional configuration item be added either to `preferences` or as a separate file called `paths` that lists a set of starting points fed to `CreateSnapshotFromDirectory()` instead of just the top.  The code that populates the list of directories and skipped files would iterate over that list instead of just `top` and should get the desired result.

If the file is absent, Duplicacy should behave as it does now, starting in the directory above the configuration.

This should leave the filters to do their job and prevents [having to search the entire system for files to be backed up](https://github.com/gilbertchen/duplicacy/issues/216#issuecomment-332965504).

tbain98:
Wouldn't the proposal in #217 give exactly what you're looking for here (the ability to change the directories being backed up to one or more directories that are not necessarily located in the same fixed directory as the configuration)? Is there anything you're requesting here that can't be achieved by #217?

markfeit:
@tbain98 I skimmed through the entire list of open issues and don't know how I missed that one.  #217 appears to be more about multiple, separate repositories within the same file system or multiple roots in the same one where this one is about streamlining the way to select subsets without having to do filter gymnastics.  I think this one has separate merits on its own, and I'm going to leave a comment in yours.

tbain98:
I would describe it as a single repository whose contents are all managed
as a bundle but with the ability to have more than one root directory,
rather than multiple repositories. That's partly just a semantic
difference, but I think that the mental model of a repository being defined
[...]

Issue #217, opened 2017-09-28
Store the repository's root folders (plural) in the .duplicacy config file


tbain98:
Please modify Duplicacy to:

1. Store the path (preferably absolute, but I could see allowing relative as well) to the root directory of the repository within the .duplicacy file. This would allow there to be no Duplicacy-related content written within the repository, and would allow multiple repositories to be defined based on different children of the same parent folder. (e.g. Repo1 includes /var/foo1 and /var/bar2 and Repo2 includes /var/foo2 and /var/bar1, by defining both to have /var as their root directory and then specifying includes/excludes appropriately.)
2. Allow multiple root directories to define a repository collectively. The current approach of storing the .duplicacy file/directory within the root directory forces us to have only a single root folder, which is limiting.

There have been comments that touched on this idea in #148, #149, and possibly others, but it's never been requested as a standalone request.

markfeit:
> Store the path (preferably absolute, but I could see allowing relative as well) to the root directory of the repository within the .duplicacy file.

I'm reading this to mean inside of `.duplicacy/preferences`, which I think would be a good thing because Duplicacy is hung up on the current working directory being the top of the tree to be backed up.

 > This would allow there to be no Duplicacy-related content written within the repository...

You can keep the Duplicacy content elsewhere by pointing to a different directory in the `.duplicacy` file...
```
# cat /foo/.duplicacy
/bar/duplicacy-config

# ls /bar/duplicacy-config
cache  filters  preferences
```
...or by using the `-pref-dir` switch to force configuration and cache to go elsewhere when the CLI is invoked.

 > ... and would allow multiple repositories to be defined based on different children of the same parent folder.

You can do this now by running `backup` multiple times with a different `-pref-dir` for each repository.  

 > Allow multiple root directories to define a repository collectively.

That can get hairy.  If you have roots `/foo` and `/bar` in the same storage under the same ID and both contain `baz`, there has to be information stored about which root `baz` came from (not unusual since most Unix file systems create a `lost+found` in their root directories).  That would add another layer of identification underneath the `id` already in the configuration.  At that point, you'd be better off doing multiple configurations or making `/` the root and including `foo` and `bar`.

tbain98:
GitHub is almost certainly going to be unhappy with the formatting of this
content. I apologize in advance, and I'll plan to clean it up from a
computer later.

Store the path (preferably absolute, but I could see allowing relative as
well) to the root directory of the repository within the .duplicacy file.

[...]

Issue #149, opened 2017-09-03
backup command should take path to backup


robbat2:
Presently, if you want to backup a specific path, you have to do `cd $DIR && duplicacy backup`. This leaves extra file descriptors temporarily open in the directory and makes some scripting more painful.

`duplicacy backup` should allow passing a SINGLE path as the location to back up. See also my prior ticket about explicitly passing a `.duplicacy` directory to all commands.

Issue #148, opened 2017-09-03
All commands should take allow option of path to preferences directory.


robbat2:
If backing up an entire system, having to have the `.duplicacy` directories or files scattered everywhere that you are backing up is very messy, and some of the locations might not permit it (eg backing up all of /home when each homedir is mounted by autofs dynamically, and writing to /home itself is disallowed).

ALL of the commands should support a command-line option to the specific `.duplicacy` file or directory to use (and should make it ignoring the existence of any `.duplicacy` files locally).

jbrodriguez:
This is something I mentioned [here](https://github.com/gilbertchen/duplicacy/issues/132#issuecomment-325848185) too, and @gilbertchen replied he would add --includes/--excludes from the command line

tbain98:
+1

I'm intending to use Duplicacy as only one of the tools performing backups on my data (I don't want to be reliant on only one piece of software, any more than I want to be reliant on any one storage provider), and I hate the idea that each of them will force me to put their "droppings" within the content to be backed up. Being able to externalize that configuration to another location sounds like a positive improvement to me.

tbain98:
@jbrodriguez What @robbat2 is proposing here is a little different from what you're advocating for in that thread. I think there's value to having a directory that contains configuration settings for the content to be backed up (whereas you appeared to be advocating a settings-free approach), but I'm 100% behind @robbat2's suggestion that the settings directory not having to be within the content to be backed up. This would also facilitate the use case you were asking for where you've got multiple backup sets defined from different subsets of the contents of a given folder, which Duplicacy currently can't support because there can be only one `.duplicacy` file for a given root folder.

jbrodriguez:
@tbain98, yes I like the idea of specifying a ```.duplicacy``` from the command line, but you should be able to set filters to add folders from anywhere in the file system (any volume, any subfolder) to your 'backup set'.

I'm just not so sure that this is so explicit (possible?) right now.

jsreynolds:
Am I missing something, or can't you just use `--pref-dir` during the init and place all of the backup configuration settings somewhere else entirely, and/or back them up separately if so desired? You can even move them later rather easily too.

robbat2:
@jsreynolds If you pass `--pref-dir`, it still drops a `.duplicacy` file, containing the path to  the `pref-dir`. I don't want that file to exist at all (and instead I will pass the path to the `pref-dir` via CLI arg and/or environment).

[...]

PR #396, opened 2018-04-08
Store repo-path in config, select repo as command argument

fracai:
This one covers fully or partially a few issues: #148, #149, #217, #302 

To `init` I've add:
 - `--repository-dir` to manually specify the repository root
 - `--no-pref-link` to disable dropping the `.duplicacy` file even if `--pref-dir` is used to change the configuration directory.

That last one also means you'd need to specify the path to the pref-dir with every command. That means that `--pref-dir` is now a global option.

This also means that the repository root is now stored in the preference file. It doesn't implement multiple repository roots, but it does allow backing up different data sets without switching directories. I intend to use this to keep all my configurations in a single directory (ie. ~/.config/duplicacy/backup_set_N).

I've done some testing, but I wouldn't be surprised if I haven't covered every use case. Corrections, comments, and test cases are welcome.

### DIFF
--- a/duplicacy/duplicacy_main.go
+++ b/duplicacy/duplicacy_main.go
@@ -38,10 +38,14 @@ func getRepositoryPreference(context *cli.Context, storageName string) (reposito
 
     var pref_dir string = context.GlobalString("pref-dir")
     if pref_dir == "" {
-        repository, err := os.Getwd()
-        if err != nil {
-            duplicacy.LOG_ERROR("REPOSITORY_PATH", "Failed to retrieve the current working directory: %v", err)
-            return "", nil
+        repository = context.String("repository-dir")
+	    if repository == "" {
+	        var err error = nil // golang complains err is undefined if it isn't defined here
+		    repository, err = os.Getwd()
+            if err != nil {
+                duplicacy.LOG_ERROR("REPOSITORY_PATH", "Failed to retrieve the current working directory: %v", err)
+                return "", nil
+            }
         }
 
         for {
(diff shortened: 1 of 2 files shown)


duplicacy/duplicacy_main.go
```diff
@@ -38,10 +38,14 @@ func getRepositoryPreference(context *cli.Context, storageName string) (reposito
...
     var pref_dir string = context.GlobalString("pref-dir")
     if pref_dir == "" {
-        repository, err := os.Getwd()
-        if err != nil {
-            duplicacy.LOG_ERROR("REPOSITORY_PATH", "Failed to retrieve the current working directory: %v", err)
-            return "", nil
+        repository = context.String("repository-dir")
+	    if repository == "" {
+	        var err error = nil // golang complains err is undefined if it isn't defined here
+		    repository, err = os.Getwd()
```
TheBestPessimist: Golang complains because you should've written `    repository, err = os.Getwd()`. I think you missed the `:` in `:=`

fracai: Gah, yes. Thank you for noticing this. I must have dropped that during a copy/paste.

gilbertchen:
I just checked in https://github.com/gilbertchen/duplicacy/commit/72239a31c42dd6e01521be19f45d13ed9caf635f which is based on your work, but without the global `--pref-dir` option.

The `--pref-dir` option isn't needed because it can always find the `.duplicacy\preferences` file under the current working directory.  Once the preference is loaded, the path stored in the `RepositoryPath` is taken to become the repository instead of the current working directory.

fracai:
Part of the advantage of `--pref-dir` is that it allows running the script from an arbitrary location. So it'll change my scripts (they'll need to `cd` to the correct location first), but it'll still work.
Thanks.

PR #405, opened 2018-04-16
enum command: useful for debugging filter patterns

SWB2:
This implements a new command, `enum`, which provides functionality useful for debugging filter patterns.

This command essentially executes just `CreateSnapshotFromDirectory()` (albeit a new version called `CreateSnapshotFromDirectoryFilterDebug()`) on the repository directory, but provides additional options for logging the snapshot creation process. For example:

* You can list entries included or excluded from the snapshot by the active filter patterns.
* You can list entries explicitly affected by a filter pattern or those implicitly processed by default.
* You can list files, directories, or both.

The output prints whether the entry is included or excluded in the snapshot, the rule (if any) that caused the entry to be included or excluded, and the full path of the entry. Output is tab-delimited, for ease of importing into a spreadsheet for further analysis.

    NAME:
       duplicacy enum - enumerate the the repository
    
    USAGE:
       duplicacy enum [command options]
    
    OPTIONS:
       -all                 DEFAULT: include any file or directory (equivalent to -explicit -implicit)
       -explicit            include files or directories affected by a pattern in the filters file (explicit processing)
       -implicit            include files or directories not affected by any pattern in the filters file (implicit processing)
       -entries, -e         DEFAULT: list files and directories (equivalent to -ientries -xentries)
       -ientries, -ie       list included files and directories (equivalent to -ifiles -idirs)
       -xentries, -xe       list excluded files and directories (equivalent to -xfiles -xdirs)
       -files, -f           list files (equivalent to -ifiles -xfiles)
       -ifiles, -if         list included files
       -xfiles, -xf         list excluded files
       -dirs, -d            list directories (equivalent to -idirs -xdirs)
       -idirs, -id          list included directories
       -xdirs, -xd          list excluded directories

### DIFF
--- a/duplicacy/duplicacy_main.go
+++ b/duplicacy/duplicacy_main.go
@@ -885,7 +885,7 @@ func listFiles(context *cli.Context) {
 		return
 	}
     
-    _, _, _, err = duplicacy.CreateSnapshotFromDirectoryFiterDebug("", top, filterDebugOptions)
+    _, _, _, err = duplicacy.CreateSnapshotFromDirectoryFilterDebug("", top, filterDebugOptions)
     
 	if err != nil {
         duplicacy.LOG_ERROR("SNAPSHOT_LIST", "Failed to list the directory %s: %v", top, err)
--- a/duplicacy/duplicacy_main.go
+++ b/duplicacy/duplicacy_main.go
@@ -792,6 +792,109 @@ func restoreRepository(context *cli.Context) {
 	runScript(context, preference.Name, "post")
 }
 
+func listFiles(context *cli.Context) {
+	setGlobalOptions(context)
+	defer duplicacy.CatchLogException()
+
+	if len(context.Args()) != 0 {
+		fmt.Fprintf(context.App.Writer, "The %s command requires no arguments.\n\n", context.Command.Name)
+		cli.ShowCommandHelp(context, context.Command.Name)
+		os.Exit(ArgumentExitCode)
+	}
+
+    filterDebugOptions := &duplicacy.FilterDebugOptions {
+         ListFilteredEntries     : false,
+         ListUnfilteredEntries   : false,
+         ListIncludedFiles       : false,
+         ListIncludedDirectories : false,
+         ListExcludedFiles       : false,
+         ListExcludedDirectories : false,
+    }
+    
+    if context.Bool("explicit") {
+        filterDebugOptions.ListFilteredEntries     = true;
+    }
+    
+    if context.Bool("implicit") {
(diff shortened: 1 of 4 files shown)


TheBestPessimist:
This is quite nice! (didn't look at the implementation, just tried `duplicacy enum > a.txt` and checked the output.

May i complain that the name `enum` is not very clear about what the command actually does?
Could `test-filters` or `check-filters` be  a better name?

SWB2:
> the name `enum` is not very clear about what the command actually does … could `test-filters` or `check-filters` be a better name?

Well, it _enumerates_ the repository, so it _sort-of_ does what it says on the tin. Having said that, I'm not at all married to the name. The functionality is all I was after. I really wanted to call it `list`, but that was already taken. `enum` was the best alternative I was able to think up. `test-filters` or `check-filters` are both perfectly reasonable.

> by just running `duplicacy enum` it seems that the folders are not listed in the same order as when running the `backup` command

All I do is insert logging output into the normal `ListEntries()` function, which is called by `CreateSnapshotFromDirectory()`, which in turn is used by `Backup()`. The output is therefore listed in exactly the same order that Duplicacy processes it, which might be helpful in diagnosing filter precedence issues.

That initially tripped me up as well: I couldn't find a directory mentioned in the output, for example, that I thought I had made a filter for. It turns out it was listed much earlier than where I was looking, which revealed that Duplicacy was first applying filters to everything in a directory (including subdirectory names) before descending into subdirectories.

> just to be sure: by default `enum` is run as `duplicacy enum -all -entries` right?

Correct.

TheBestPessimist:
Also: you already helped me fix a bug with my ignore file :D

SWB2:
> you already helped me fix a bug with my ignore file :D

Great! I am using your file as a base for my own. :-)

TheBestPessimist:
@gilbertchen what's your opinion on the naming  `enum` vs `test-filters` vs `check-filters` vs < `other` >?
I this this should be changed before the pull request acceptance.

gilbertchen:
Instead of adding an `enum` command, I added an `-enum-only` option to the `backup` command in commit https://github.com/gilbertchen/duplicacy/commit/aadd2aa390355b21a2d986da15457c211900b941.  I think this is simpler as you can see from the code changes.  Moreover, the `enum` command would create two separate code paths for listing the repository which makes it less dependable for debugging purposes.

Those options for the `enum` command are very useful, but you can emulate them by running `duplicacy -d -log backup -enum-only` and then filtering the output with `grep`.

gilbertchen:

https://forum.duplicacy.com/t/backup-command-details/1077/1

Issue #341, opened 2018-01-26
macOS High Sierra APFS Snapshot Support


leftytennis:
@gilbertchen, I'd be happy to take a stab at trying to add snapshot support for Apple's new High Sierra filesystem called APFS.

Could you give me a few pointers which source files I'd need to implement this functionality?

I'm assuming generically that I'd need to minimally do the following:

1. Add new backup command line switch like -apfs (similar to -vss)
2. If -apfs is specified, verify that the underlying filesystem is Apple's APFS
3. Create a snapshot of that volume
4. Mount the newly created snapshot
5. Backup using the newly mounted snapshot as the source instead of the original source volume
6. Unmount the snapshot
7. Delete the snapshot

If you could provide some hints which src files I'd have to review/change, it would help me w/o having to just search all of the source, etc.

Thanks

Edit: A little googling shows that APFS snapshot requires that a program have a special entitlement and may also require root access. Not clear how this special entitlement is requested or granted by Apple, so perhaps this isn't quite ready for prime time yet.

gilbertchen:
This is great!  The VSS is one of the few things that Windows does better than Linux IMHO and an APFS equivalent would benefit many macOS users.

To start with, the -vss option is checked in `duplicacy/duplicacy_main.go` and passed to `duplicacy.Backup()`, which is in `src/duplicacy_backupmanager.go`.  The only relevant part there is:
https://github.com/gilbertchen/duplicacy/blob/ffe04d691bb866795bfe69e8d20f48c51e3bfaaa/src/duplicacy_backupmanager.go#L183-L184

So you just need to implement two functions `CreateShowCopy()` and `DeleteShowCopy()`.  `CreateShadowCopy()` return the alternative path of the repository that Duplicacy need to access the created shadow copy.  Currently these 2 functions are only implemented for Windows.

I don't know how you can call the APFS library from go, but if cgo is required, you can assume that XCode is installed since it is required by the installation: https://github.com/gilbertchen/duplicacy/wiki/Installation#build-on-macos

fracai:
I was looking in to this one and in order to create the snapshots from duplicacy natively, you'd need to request a specific entitlement (I think it is `com.apple.developer.vfs.snapshot`) to be added during code signing; it looks like duplicacy already is signed. That needs to be requested from Apple Developer Tech Support. I can't find anything about that specifically, but it seems that it has been granted to backup applications after review.

[...]

PR #415, opened 2018-04-28
Add APFS snapshot support

amarcu5:
Add's VSS support for macOS using APFS snapshot

Addresses issue #341


jt70471:
I haven't patched my copy yet, but will do so and test. Thanks for doing this, too bad Apple has decided to make the API for APFS snapshots so convoluted to use. The API itself seems simple, minus the criterial for the special entitlement...

amarcu5:
@jt70471 All done, thanks, I'd forgotten about that

amarcu5:
@jt70471 agreed, keep the suggestions coming!

While I was at it I noticed I inadvertently broke Windows builds but all fixed now

jt70471:
This looks great, the only thing I couldn't test was the check that the filesystem is indeed APFS, because I have already converted all of my macs to APFS!

gilbertchen:
Thanks for your contribution and testing!  I can't test it myself but if it has been running good for you guys that it should be safe to merge it.

Issue #143, opened 2017-08-30
support for metadata, Mac extended attributes, etc.


fracai:
I was originally just concerned with MacOS resource forks, but was reminded of the [Backup Bouncer](https://github.com/n8gray/Backup-Bouncer) test suite.

Macs currently split non-data fork data to AppleDouble encoded "**._filename**" files when saving to non-Mac systems.

These forks can be listed and displayed with the _xattr_ command.

The easiest way to create a file with a resource fork is dragging a URL or text selection to the desktop to create a "**webloc**" or "**textClipping**" file.

gilbertchen:
Resource forks are backed up as extended attributes.  However, when they are restored on an OS without resource fork support, "._filname" files will not be created.

fracai:
Yeah, I was looking through the code earlier and realized you were indeed looking at extended attributes. When listing a backup I saw a resource fork file listed with zero size and assumed attributes were being skipped. I'll going to give a try at running Backup Bouncer.

fracai:
I created [scripts](https://github.com/fracai/Backup-Bouncer/tree/master/scripts) for testing rsync and Duplicacy against BackupBouncer. Please let me know if I have duplicacy configured incorrectly as in my tests I couldn't get resource forks to be restored. BackupBouncer reports a number of failures, but I'm primarily interested in resource forks.

[bbouncer-duplicacy.txt](https://github.com/gilbertchen/duplicacy/files/1265700/bbouncer-duplicacy.txt)

gilbertchen:
This is a bug.  The attributes are discarded if there are no patterns provided to the restore command.  But if there are (which means you're restore one file or a subset of files), attributes are restored correctly.

This bug has been fixed by https://github.com/gilbertchen/duplicacy/commit/dfa61132794762cad5574a686f6691fa67136d81

sharpjs:
I'd really like to see the post-fix Backup Bouncer results.  That would save me the time of doing them myself. 😄   Duplicacy is on my short list for a CrashPlan replacement, but it's a non-starter if the results are anything like that posted above.

fracai:
I checked out the latest source, but I haven't had a chance to build or test yet. Hopefully soon. I'll certainly post here once I have. I'm in the same shot. Duplicacy is the current leader.

fracai:
[...]

PR #417, opened 2018-04-29
Fixed restoration of basic UNIX file permissions

amarcu5:
Fixed restoration of setgid/setuid bits (see issue #143)

Now passes Backup-Bouncer's critical basic permissions test


gilbertchen:
This looks good.  Thanks for the fix!

Issue #435, opened 2018-06-04
Why are size stats units given as thousands?


fracai:
This is a relatively minor point, but why are the size stats (duplicacy check -stats) given in thousands of units? As in, 1 GB of data is presented as 1,000M. 1 MB is presented as 1,000K. Even knowing what is being shown, I still need to make two conversions in my head. First, pretend that the comma is really a decimal. Then, increase the units notation.

The fix would be a small change to the `PrettyNumber` and `PrettySize` methods in `duplicacy_utils.go`, but I figure there's probably a reason for the current appearance. Would this be a welcome pull request or is the current behavior explicitly preferred?

gilbertchen:
It was a mistake to use 1000 instead of 1024.  Please create a pull request if you can.

fracai:
To clarify, it's no so much 1000 vs 1024 that I was talking about (though I can fix that too).
I was talking about how sizes are reported like:
`123,456 M`
instead of:
`123.456 G`
or, with the 1000 to 1024 fix:
`120.563 G`

It's mostly the units that I'm questioning.

NiJoao:
I agree that 3 places before and after a comma is clearer, don't need to go into the thousands units.
But please, please, don't use MB/GB in powers of 1024. Always use SI units, MB with 1000 or MiB with 1024, https://en.m.wikipedia.org/wiki/Mebibyte

fracai:
I created a pull request that illustrates what I'm going for.
It also uses the SI units.

PR #437, opened 2018-06-04
Standardize scale of stats units

fracai:
reformat results of PrettyNumber and added GiB range to PrettySize
added SI suffixes to each

as discussed in #435

### DIFF
--- a/src/duplicacy_backupmanager.go
+++ b/src/duplicacy_backupmanager.go
@@ -483,7 +483,7 @@ func (manager *BackupManager) Backup(top string, quickMode bool, threads int, ta
 					remainingTime = (totalModifiedFileSize-uploadedModifiedFileSize)/speed + 1
 				}
 				percentage := float32(uploadedModifiedFileSize * 1000 / totalModifiedFileSize)
-				LOG_INFO("UPLOAD_PROGRESS", "%s chunk %d size %d, %s/s %s %.1f%%", action, chunkIndex,
+				LOG_INFO("UPLOAD_PROGRESS", "%s chunk %d size %d, %sB/s %s %.1f%%", action, chunkIndex,
 					chunkSize, PrettySize(speed), PrettyTime(remainingTime), percentage/10)
 			}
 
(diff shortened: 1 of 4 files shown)


fracai:
I note that PrettySize and PrettyNumber are very similar. Is it worth combining them? Or should they stay separate with their differences?

fracai:
Good point, I've removed the 'B' suffix from each. I looked at each of their uses and I think it makes sense now. Let me know if you want any adjustments. Specifically, an alternate change would be to put the 'B' back in and adjust each use of the methods to remove the redundant 'bytes'.

jt70471:
I have not applied this patch yet, but seems the numbers will be left with K/M/G/T or Ki/Mi/Gi/Ti suffixes... shouldn't they all end with B? i.e. KiB, MiB, GiB, TiB, etc?

fracai:
Yes, @gilbertchen suggested that the 'B' could be problematic given that other spots in the code supply 'bytes'. I'm thinking I should take another look at this to add the 'B' back in and remove 'bytes' where applicable from the code.

jt70471:
The standard is KiB and KB, etc. so I think it would be best to stick to the standard and remove "bytes" anywhere is it found after the PrettyNumber or PrettySize formatted number.

fracai:
Yeah, that's what I'm headed for

fracai:
The additional commit restores the "B" and reformats the log messages so there isn't any duplicate "bytes" phrasing.

gilbertchen:
I don't like separating `B` and `ytes` in different locations.  Format strings like ``Files: %d total, %sytes; %d new, %sytes` look awkward.

May I suggest using K/M/G/T only and then the format strings can use either `%sB` or `%s bytes`?

fracai:
yes, I'll make that change in a few hours.

fracai:
I think I've covered all the relevant format strings. I moved the "B" out of the PrettyNumber and PrettySize methods and put "B" in the relevant format strings.

jt70471:
Thanks for all of your work, but I really don't like MiBytes, GiBytes, etc.. Never seen the Bytes written out like that in any other usage, I would prefer just plain old MB, MiB, GB, GiB, etc.

fracai:
I'm fairly neutral with how the units are expressed. At this point I think it's whatever @gilbertchen wants. :-)

gilbertchen:
I'm closing this PR without merging it.  The notions of Ki, Mi, Gi, etc would be in conflict with the input where we use `-c 1M` or `-c 128K` to specify the chunk size in the `init` and `add` command.  Although technically KiB, MiB and GiB are more correct, it seems K, M, G are ok and generally acceptable.  Therefore I want to leave these as they are now.

PR #441, opened 2018-06-05
Implement multithreaded pruning

gilbertchen:
With this change you can now pass a `-threads` option to the prune command to use any number of threads to delete or rename chunks concurrently.

### DIFF
--- a/duplicacy/duplicacy_main.go
+++ b/duplicacy/duplicacy_main.go
@@ -1066,12 +1066,17 @@ func pruneSnapshots(context *cli.Context) {
 		os.Exit(ArgumentExitCode)
 	}
 
+	threads := context.Int("threads")
+	if threads < 1 {
+		threads = 1
+	}
+
 	repository, preference := getRepositoryPreference(context, "")
 
 	runScript(context, preference.Name, "pre")
 
 	duplicacy.LOG_INFO("STORAGE_SET", "Storage set to %s", preference.StorageURL)
-	storage := duplicacy.CreateStorage(*preference, false, 1)
+	storage := duplicacy.CreateStorage(*preference, false, threads)
 	if storage == nil {
 		return
 	}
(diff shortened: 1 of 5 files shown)


TheBestPessimist:
If my understanding from a10 minute code review is that what you did with ChunkOperator is like a queue of commands which are dispatched to waiting threads. Am i correct?

TheBestPessimist:
May i also suggest that you add some tests which use more than 1 thread? Right now i see that all tests are only using 1.

gilbertchen:
It was manually tested before, but now I added a new test named `TestPruneMultipleThread`.

TheBestPessimist:
From the implementation i have the feeling that it may not actually be a problem, but i will also ask: if i run multithreaded prune from my 3 machines against the same storage there' won't be any problems right?

no exhaustive or exclusive just the normal `duplicacy prune -threads 20 -keep 7:180 1:7`

bassebaba:
Will this speed up pruning?
I have a 3TB backup set in which I do pruning once a month which removes about 0.5TB data.
This takes like 12 hrs each time :P

The connection between the sites are 100/100Mbit and there's no problem with utilizing that connection.
The pruning uses around 2 megabytes / sec upstream from the backup location.

For me it seems INSANE that the prune uses 2 megs / sec upstream from the backup destination for 12 hrs!
How does this work? Why is it so data-intense to do the prune?

12hrs *60 min *60 sec *2 megs = 86400 megabytes ~= 86 gigs.
 
So euuhh. Euhm. I need to transfer 86 gigs in order to prune about 500 gigs?

Please explain this to me.

It seems to me that we really could either use a server-side here, or some sort of improvement to the "algo", since it's insane amounts of data that needs to be transferred to the client to do the pruning....

gilbertchen:
@TheBestPessimist it won't corrupt the backups but on some machines the prune command may error out if multiple of them happen to rename the same chunk.

@bassebaba Yes, this will definitely speed up pruning.  I don't know where 2 megabytes /sec comes from, but the prune command won't use that much traffic -- it is just one or two HTTPS requests for each chunk, and these requests will be very small since the chunk itself doesn't need to be uploaded or downloaded.

TheBestPessimist:
By "error out" you mean that the actual prune will just stop at the first error, or that it will show errors for some chunks, but continue until prune is finished

Issue #443, opened 2018-06-07
Copy to SFTP fails


NiJoao:
Hi,
I don't seem to be able to copy my 700GB local repository to an SFTP one.

Local Chunk folder:

Chunk folder on the SFTP:
149 GB, 45 533 Files, 256 Folders. (copied via SFTP client, same error when empty)

Steps made on the local machine:
```
mkdir -p Downloads\repo
cd C:\Users\USER\Downloads\repo

.duplicacy\duplicacy_win_x64_2.1.0.exe init -e -c 4M repo D:\Partilha\Backups\Duplicacy
## The repository already existed, used same password, everything succeeded.
.duplicacy\duplicacy_win_x64_2.1.0.exe add -copy default -e -bit-identical offsite_storage repo sftp://USER@SERVER:PORT/Duplicacy
## Succeedes, creates the fodlers and config files on the server.

.duplicacy\duplicacy_win_x64_2.1.0.exe list -all
## This succeeds, lists thousands of chunks.
.duplicacy\duplicacy_win_x64_2.1.0.exe list -all -storage offsite
## This succeeds, listing nothing.

.duplicacy\duplicacy_win_x64_2.1.0.exe -d copy -threads 1 -to offsite -upload-limit-rate 800 -id Some_ID -r 1
```

The last command outputs Hundreds of lines "Listing chunks", but ends with:

```
....
Listing chunks/02/
Listing chunks/01/
Listing chunks/00/

Found 30618 chunks on destination storage
Chunks to copy = 215, to skip = 53, total = 268
Total chunks in source snapshot revisions = 268

Copying chunk 91a618e10a617207cf52cf040a994cde40d73ef3c4025370a170d0fee1905be2 to 91a618e10a617207cf52cf040a994cde40d73ef3c4025370a170d0fee1905be2
Chunk 91a618e10a617207cf52cf040a994cde40d73ef3c4025370a170d0fee1905be2 has been downloaded
Copying chunk ee9f5c4dbf9a75b3da95308413243c5b22ed321212e07069dbb5c026f73f8ec7 to ee9f5c4dbf9a75b3da95308413243c5b22ed321212e07069dbb5c026f73f8ec7
Fetching chunk ee9f5c4dbf9a75b3da95308413243c5b22ed321212e07069dbb5c026f73f8ec7
Chunk ee9f5c4dbf9a75b3da95308413243c5b22ed321212e07069dbb5c026f73f8ec7 has been downloaded
[...]

PR #447, opened 2018-06-13
Acknowledge malware/spam warnings from GCD

s4y:
If Google thinks that a file is malware or spam (which can happen
spuriously to blobs of encrypted data), it will prevent the initial
download and return an error with reason "cannotDownloadAbusiveFile".
The API expects a program to prompt the user in this case and then,
optionally, let them bypass it.

Ideally duplicacy should prompt, but this patch just logs a warning.

When I printed `err.(*googleapi.Error)`, its `Errors` field was empty,
hence the sketchy string matching. It's possible that I did something
wrong, though.

Fixes issue #443.

### DIFF
--- a/src/duplicacy_gcdstorage.go
+++ b/src/duplicacy_gcdstorage.go
@@ -624,13 +624,22 @@ func (storage *GCDStorage) DownloadFile(threadIndex int, filePath string, chunk 
 	var response *http.Response
 
 	for {
-		response, err = storage.service.Files.Get(fileID).Download()
-		if retry, err := storage.shouldRetry(threadIndex, err); err == nil && !retry {
+		// AcknowledgeAbuse(true) lets the download proceed even if GCD thinks that it contains malware.
+		// TODO: Should this prompt the user or log a warning?
+		req := storage.service.Files.Get(fileID)
+		if e, ok := err.(*googleapi.Error); ok {
+			if strings.Contains(err.Error(), "cannotDownloadAbusiveFile") || len(e.Errors) > 0 && e.Errors[0].Reason == "cannotDownloadAbusiveFile" {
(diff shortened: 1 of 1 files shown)


jt70471:
This should be a command line switch in my opinion to leave it to the user whether they want to download potentially dangerous content.

gilbertchen:
Do you still keep the chunk that caused the error?  If so do you mind sharing the file with me so I can reproduce the error myself?

s4y:
@gilbertchen Sure thing. I made it public under the assumption that the chunk is encrypted and doesn't leak any information on its own, but let me know if that's not the case and I'll lock it down:

https://drive.google.com/file/d/1BAn6XfBFWTjmH5Vjn7OgW25pOo-GKuzZ/view

TowerBR:
@s4y , I tried to download the file out of curiosity, but ...

> "Sorry, this file is infected with a virus
> Only the owner is allowed to download infected files."

Have you tried uploading the file to www.virustotal.com to see what the various antivirus programs say?

gilbertchen:
I was able to download the file by creating a copy in Google Drive and then downloading the copy.

virustotal.com results: https://www.virustotal.com/#/file/a99631f46c2e2458cc0b6c309dcc8bac64ada55422e676f172b78d49ee5e8b91/detection

s4y:
And, just to make sure, the file does look like the expected Duplicacy encrypted noise, right?

gilbertchen:
@s4y yes, that chunk file is encrypted by a randomly generated key stored in your `config` file so no one else can decrypt it.

gilbertchen:
Thank you for the fix.  The string matching is fine as it looks like the Google Drive library code doesn't unmarshal the body before returning the error.

Issue #425, opened 2018-05-11
URL request '...' returned status code 408


DreamFlasher:
On backblaze every fourth chunk yields a URL request '...' returned status code 408. Is duplicacy trying again to upload the chunk or is it gone and duplicacy believes the upload worked? (I guess the uploaded chunk debug log could/should contain the URL?)

gilbertchen:
Backblaze servers return different errors all the time, and Duplicacy always tries to re-upload the chunk on any errors.  I wonder why it happens so regularly.  What does the URL look like?  Can you post it here with the credential token removed?

DreamFlasher:
Potentially identifying information randomly changed: `URL request 'https://pod-000-1084-19.backblaze.com/b2api/v1/b2_upload_file/daab65d8c4031bbf2a3e0711/c002_v0001083_t0047' returned status code 408`
As I didn't manage to get a full backup run through and the hashing starts from scratch every time, I changed the source code to have the fallback loop run infinitely -- now it works like a charm.

Issue #423, opened 2018-05-08
Turn off Maximum backoff reached limit or expose as option


DreamFlasher:
A flaky internet connection currenly breaks backups. 
Please either turn off the maximum backoff limit or allow the user to set it higher/turn it off. 
`Failed to upload the chunk *: Maximum backoff reached`

TheBestPessimist:
I vote for providing an option for changing the number of retries. 
This could be either per storage or global setting (using `set` command?)

PR #462, opened 2018-07-17
Too low backoff limit breaks backup with B2

DreamFlasher:
Fixes https://github.com/gilbertchen/duplicacy/issues/425

### DIFF
--- a/src/duplicacy_b2client.go
+++ b/src/duplicacy_b2client.go
@@ -97,7 +97,7 @@ func (client *B2Client) call(url string, method string, requestHeaders map[strin
 	var response *http.Response
 
 	backoff := 0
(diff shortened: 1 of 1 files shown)


gilbertchen:
This won't work -- the backoff mechanism should be modified for such a high number of retries.

DreamFlasher:
Well it works for me, as described in #425. Can you at least make the backoff limit configurable? I always have to patch and rebuild the code myself without that.

TheBestPessimist:
@DreamFlasher you already requested that in #423  (and i totally approve!)

DreamFlasher:
@TheBestPessimist Thanks. As issue #423 was never adressed but instead my PR here was closed I figured I write it again.

PR #475, opened 2018-08-02
b2client: Add the ability to use a restricted Application ID

### LINKED ISSUE #484: Support for B2 Application Keys
Backblaze recently made B2 Application Keys available, a feature that lets you create application specific keys, and which also can restrict access to specific buckets. Using the B2 account for more than one thing, it would be very nice to use they more restricted keys. I tried using a generated API key (restricted to one only one bucket), but it doesn't seem to work with the current version - 2.1.1:

```
chrille@volgograd $ duplicacy init -e data-backup b2://***
Enter Backblaze Account ID: app-key-id
Enter Backblaze Application Key: app-key
URL request 'https://api002.backblazeb2.com/b2api/v1/b2_list_buckets' returned status code 400
URL request 'https://api002.backblazeb2.com/b2api/v1/b2_list_buckets' returned status code 400
...
```

* [Official Documentation](https://www.backblaze.com/b2/docs/application_keys.html?_ga=2.44038181.1038729054.1534753353-646273881.1534753353)
* [Announcment in Blog](https://www.backblaze.com/blog/b2-application-keys/)

bekriebel:
B2 allows you to create Application IDs that are restricted to specific
buckets. To allow for this, read back the information provided by the
authorization request and use the allowed bucket permissions if they are
found.

An account-level Application ID will still continue to function
normally, as would an Application ID that has access to all buckets.

Fixes #484

### DIFF
--- a/src/duplicacy_b2client.go
+++ b/src/duplicacy_b2client.go
@@ -40,6 +40,7 @@ var B2AuthorizationURL = "https://api.backblazeb2.com/b2api/v1/b2_authorize_acco
 type B2Client struct {
 	HTTPClient         *http.Client
 	AccountID          string
+	AllowedBucketID    string
 	ApplicationKey     string
 	AuthorizationToken string
 	APIURL             string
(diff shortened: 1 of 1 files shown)


gilbertchen:
I just pushed commit https://github.com/gilbertchen/duplicacy/commit/674d35e5cae603b641a048609ff48335a36ebb36 which is basically half of the changes in this PR.  This other half (`AllowedBucketID`) isn't needed as Backblaze has fixed an issue in `b2_list_buckets` when an application key is used.

Thanks for your work!

PR #479, opened 2018-08-07
Small changes - labels of the preferences file

TowerBR:
Small changes in the structure labels of the preferences file to make them clearer.


TheBestPessimist:
I don't think it's going to be that simple. 
Right now, if i use your branch all my repositories are dead.

What you should do is also provide a means of migration which changes only those particular strings and leaves everything else intact. (this is because you cannot guarantee what changes the user did to the preferences file either by using the `set` command out by manual edits)

TowerBR:
Hmm, I'll evaluate better / deeper ...

But the idea is proposed ... ;-)

gilbertchen:
Closing this one as it will break all existing preferences files.

PR #495, opened 2018-09-18
Replace special characters in environment variable name with underscores

plasticrake:
If the storage name contains a special character, and the first check for an environment variable does not return a value, this change does an additional check looking for an environment variable with all special characters replaced with underscores.

The duplicacy documentation references example storage names with hyphens such as `webdav-storage` but this is not valid in an environment variable name in Linux/Mac (in most shells). 

With this change duplicacy will look for:
`DUPLICACY_WEBDAV-STORAGE_PASSWORD`
and if not found also look for:
`DUPLLICACY_WEBDAV_STORAGE_PASSWORD`


gilbertchen:
This is great.  Thanks!

PR #496, opened 2018-09-20
Change in prune message

TowerBR:
This PR was created based on Christoph's suggestion in this forum post:(https://forum.duplicacy.com/t/naming-repository-id-or-snapshot-id/1113/22)


gilbertchen:
The format is wrong -- the arguments need to be swapped too.

PR #498, opened 2018-09-22
Add exclude_by_attribute preference to exclude files based on xattr

### LINKED ISSUE #478: Feature Request: Honor com_apple_backup_excludeItem on MacOS
Reiterating @saspus's [forum post](https://forum.duplicacy.com/t/honoring-com-apple-backup-excludeitem-on-macos/979)

> On MacOS an extended attribute com.apple.metadata:com_apple_backup_excludeItem is used to tell backup software that the particular file does not need to be backed up. This is a more robust solution compared to separately maintained global exclusion list as the metadata is kept physically close to the the data it describes; that way applications, who have much better idea about how to manage their files, can decide what gets backed up and what is transient and shall be skipped.

> Time Machine honors that flag, as does CrashPlan and even Arq.

I currently mimic this using a pre-backup script that dynamically creates a filter, but this can be slow when run every time.

```bash
cd -P $backup/..
find . -xattrname "com.apple.metadata:com_apple_backup_excludeItem" -exec ls -dp {} \; | sed s/^\.\\//-/ > $backup/filters.metadata
cat $backup/filters.local $backup/filters.metadata > $backup/filters
```

plasticrake:
Add preference `exclude_by_attribute` (boolean). If true:
* **macOS**: will exclude all files that have xattr `com.apple.metadata:com_apple_backup_excludeItem` that contain the value `com.apple.backupd`.
* **Linux**: will excluded all files that have xattr `dulicacy_exclude`.

**Tests Included**, tested on macOS afps and Debian ext4

Fixes #478 

I named the preference "exclude_by_attribute" and not "exclude_attribute" to make it clear this doesn't affect the storage of attributes on the files.
Alternatively this could be named something like "file_exclude_attribute"

### Discussion:
* Reading Attributes
  * I see that attributes are not always read based on backups size and `DUPLICACY_ATTRIBUTE_THRESHOLD`. I didn't change the existing code to force a read of the attributes when this occurs. However I think that if `exclude_by_attribute` is true that it should read the attributes and discard them at least (or call `xattr.Getxattr` directly on the file if `discardAttributes` is true). _Based on input I can update the pull request._
* Configurable Attributes
  * I started building out other preferences to make this configurable and possibly support multiple attributes. However I was hesitant to add event more new preferences and adding more parameters to ListEntries(). I can update the pull request with input on design if requested. I think the best design would be to change the `bool` into a `map[string]string` like the Keys preference. _Based on input I can update the pull request._

### DIFF
--- a/duplicacy/duplicacy_main.go
+++ b/duplicacy/duplicacy_main.go
@@ -536,6 +536,11 @@ func setPreference(context *cli.Context) {
 	
 	newPreference.NobackupFile = context.String("nobackup-file")
 
+	triBool = context.Generic("exclude-by-attribute").(*TriBool)
+	if triBool.IsSet() {
+		newPreference.ExcludeByAttribute = triBool.IsTrue()
+	}
+
 	key := context.String("key")
 	value := context.String("value")
 
@@ -717,7 +722,7 @@ func backupRepository(context *cli.Context) {
 	uploadRateLimit := context.Int("limit-rate")
 	enumOnly := context.Bool("enum-only")
 	storage.SetRateLimits(0, uploadRateLimit)
-	backupManager := duplicacy.CreateBackupManager(preference.SnapshotID, storage, repository, password, preference.NobackupFile)
+	backupManager := duplicacy.CreateBackupManager(preference.SnapshotID, storage, repository, password, preference.NobackupFile, preference.ExcludeByAttribute)
 	duplicacy.SavePassword(*preference, "password", password)
 
 	backupManager.SetupSnapshotCache(preference.Name)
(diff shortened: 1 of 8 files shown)


gilbertchen:
Thank you for this work, but I'll leave this PR open without merging it, for two reasons:

* I plan to rework the directory scanning code and this may affect how xattrs are read and stored in the snapshot file
* I don't like hardcoding the attribute names.  Maybe a better way is to specify these names in the `filters` file to allow for some flexibility.

plasticrake:
@gilbertchen Is it still possible to have this feature added? I can rework this PR if needed (has the directory scanning code you referenced above been updated yet?). I've been running an old version of duplicacy using this PR and I'd like to be able to upgrade to the newest version. Thanks!

gilbertchen:

https://forum.duplicacy.com/t/honoring-com-apple-backup-excludeitem-on-macos/979/5

gilbertchen:
Thank you for your contribution.  I'm now merging this as it has been requested by other users too: https://forum.duplicacy.com/t/honoring-com-apple-backup-excludeitem-on-macos/979

gilbertchen:

shirazd:
Hi Gilbert,

I'm very happy to read that this has been merged in. My question is, do we need to do anything for this to take? eg. do we have to add a line to the filters file? Or is it automatic?

Thanks!
Shiraz

ps. I have cross-posted this question to github and the duplicacy forum

shirazd:
My question was answered on the forum:

duplicacy set -exclude-by-attribute=true

Thanks all!

gilbertchen:

https://forum.duplicacy.com/t/kopia-vs-duplicacy/6625/3

PR #510, opened 2018-10-26
Use a small window for calculating current transfer speed for backup

### LINKED ISSUE #128: Upload speed is wrong when chunks are skipped
When resuming an incomplete backup, Duplicacy takes into account the chunks that it skips as part of the uploaded data, and uses them to calculate upload speed, which results in a wrong speed and ETA information.

I would excpect Duplicacy to only take into account actually uploaded chunks.

screen shot 2017-08-26 at 08 18 30

TheBestPessimist:
With this I am trying to fix #128: the problem that upload speed is wrong when a large portion of a backup has skipped chunks (eg. 80% of backup is skipped chunks and only 20% are actually uploaded).

I have done this by implementing a circular array which stores the current uploaded MB and the timestamp when they were uploaded, and then calculates the average between the current (latest) record added and the earliest record.

Does this look like a good approach to you?

### DIFF
--- a/src/duplicacy_backupmanager.go
+++ b/src/duplicacy_backupmanager.go
@@ -386,7 +386,7 @@ func (manager *BackupManager) Backup(top string, quickMode bool, threads int, ta
 	// is the last file.
 	fileReader := CreateFileReader(shadowTop, modifiedEntries)
 
-	startUploadingTime := time.Now().Unix()
+	//startUploadingTime := time.Now().Unix()
 
 	lastUploadingTime := time.Now().Unix()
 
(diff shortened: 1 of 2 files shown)


TheBestPessimist:
Looking back at this, it seems i have quite a lot of commits. Would you be interested in merging with this many, or should i close the PR and do a new squashed one?

TheBestPessimist:
On the forums I have seen people raising issues about the wrong transfer speed, so I'm wondering: is there anything you would like to see changed @gilbertchen before merging this?

Do you maybe see this approach as wrong and would like to change it in some way??

PR #511, opened 2018-10-28
Write a log message if nothing was pruned

### LINKED ISSUE #424: Prune should write in the log when there's nothing to be pruned
From this: https://github.com/TheBestPessimist/duplicacy-utils/issues/8, when there's nothing to prune, prune should not just create an empty `prune-log-xxx` file, but actually write that "There's nothing to prune, m8". (well, a better message)

This would help newcommers understand duplicacy actions just a little bit better.

TheBestPessimist:
This should close #424.

---

This is the simplest implementation I could think of. 
There was the alternative to make a boolean and check it at every `fmt.Fprintf` but that meant i had to also modify the methods `pruneSnapshotsExhaustive` and `pruneSnapshotsNonExhaustive` which i don't think is worth it.

### DIFF
--- a/src/duplicacy_snapshotmanager.go
+++ b/src/duplicacy_snapshotmanager.go
@@ -1605,6 +1605,15 @@ func (manager *SnapshotManager) PruneSnapshots(selfID string, snapshotID string,
 
 	defer func() {
 		if logFile != nil {
+
+			// in case nothing was pruned, the log file is empty.
+			// write in the logfile that "nothing was pruned"
+			if stats, e := logFile.Stat(); e == nil {
+				if stats.Size() <= int64(2) {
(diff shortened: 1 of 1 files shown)


src/duplicacy_snapshotmanager.go
```diff
@@ -1607,6 +1605,15 @@ func (manager *SnapshotManager) PruneSnapshots(selfID string, snapshotID string,
 
 	defer func() {
 		if logFile != nil {
+
```
gilbertchen: I'm not sure if it is a good idea to call Stat() before Close().  The file may have not been actually written to the disk because the Close() call, you're likely to get a NotExist error.

TheBestPessimist: Then should i call Stat after the close, even if close gives error?

PR #512, opened 2018-11-03
Improve filter syntax with include files and change exclude behavior of .duplicacy

a-s-z-home:
Hi,

I have extended the filter file syntax with a new statement to include another filter file.
Using @<filename>, you can now include other files. Relative paths are supported.
In addition I have changed the exclude behavior for the .duplicacy folder. When using nobackup_file, this folder will not get excluded anymore. I use filters that exclude cache and internal files

### DIFF
--- a/src/duplicacy_entry.go
+++ b/src/duplicacy_entry.go
@@ -481,9 +481,6 @@ func ListEntries(top string, path string, fileList *[]*Entry, patterns []string,
 	entries := make([]*Entry, 0, 4)
 
 	for _, f := range files {
-		if (f.Name() == DUPLICACY_DIRECTORY) && (nobackupFile == "") {
-			continue
-		}
 		entry := CreateEntryFromFileInfo(f, normalizedPath)
 		if len(patterns) > 0 && !MatchPath(entry.Path, patterns) {
 			continue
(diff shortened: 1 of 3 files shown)


a-s-z-home:
I have replaced the static check for the directory completly by usage of predefined filters.
This is better than relying on nobackup_file preference.
See [forum post](https://forum.duplicacy.com/t/do-not-filter-out-duplicacy-and-change-transient-data-locations/1386/6?u=a-s-z-home)

TheBestPessimist:
May i suggest you split this kitchen-sink style PR into multiple? Right now it appears it has 2 (or were there 3?) subjects and this makes it more difficult to review/digest/accept.

TheBestPessimist:
As a sidenote: i think it's a wonderful idea to be able to include multiple filters files, as this way people can just use a generic filters file and  include inside it others (eg.  [mine](https://github.com/TheBestPessimist/duplicacy-utils/) ) without any need to modify mine. (this means easier updates!)

a-s-z-home:
OK, I can revert the changes from my master regarding the .duplicacy folder, create a new branch.
Maybe it is better to cancel this pull request and create a new one for the other branch?
Is it possible to request another pull request, that needs the changes of this pull request?
Replacing the old logic to exclude the .duplicacy folder is based on new routines.

a-s-z-home:
I am closing this pull request in favor of https://github.com/gilbertchen/duplicacy/pull/514

PR #519, opened 2018-11-20
Just a little clean-up

tempelmann:
Let me know in the comments if something isn't right.


gilbertchen:
I cherry-picked your import change (https://github.com/gilbertchen/duplicacy/pull/519/commits/e3486440a47626395d77f7ba42cdc5bf35aa16c5) into the master branch.  Other changes are fine except the .ignore file -- I generally avoid checking in that file because everyone may have their own set of excludes.

tempelmann:
Well, there are global gitignore preferences for things people may be putting into a repo dir but don't want checked in, and there there are files that should never be checked in. In this case, it's the files that are "auto-generated and not part of the actual source." You never want them in the repo, because that would be redundant and only blow up the repo. Hence the .gitignore file for that. But thanks for checking in the other changes, at least.

TheBestPessimist:
@gilbertchen i still believe you should include a generic `.gitignore` file

PR #542, opened 2019-02-15
When prefetching chunks don't stop on a chunk that may not be needed

gilbertchen:
When prefetching chunks to be downloaded during a restore operation, the current strategy is to stop at the first chunk that may not be needed, which may be too conservative.  This simple change is to keep looking at other chunks in order to prefetch enough chunks to keep the download pipeline busy.


zedkyuu:
As mentioned in the referenced issue, this adds more to the prefetch queue in the beginning, but doesn't help later on when the files are being downloaded one at a time as every file has blocks that aren't prefetched.

PR #543, opened 2019-02-18
Support downloading files from a custom public URL pointing to b2.

Mossop:
Normally downloading files from Backblaze incurs download fees. They are low but over time add up. However Backblaze has a partnership with Cloudflare where downloading files hosted at Backblaze through Cloudflare's CDN doesn't incur any costs (https://www.backblaze.com/b2/solutions/content-delivery.html)

This change adds support for a user to specify the address of the b2 bucket hosted through Cloudflare as `duplicacy set -key b2_public_url -value https://foo.example.com` and subsequently any files that need to be downloaded from backblaze use the public url making the download free.

Unfortunately I wasn't able to run the go tests (even with no code modifications) on my machine so I'm not sure if this passes them.

### DIFF
--- a/src/duplicacy_b2client.go
+++ b/src/duplicacy_b2client.go
@@ -54,11 +54,12 @@ type B2Client struct {
 	TestMode bool
 }
 
-func NewB2Client(applicationKeyID string, applicationKey string) *B2Client {
+func NewB2Client(applicationKeyID string, applicationKey string, publicURL string) *B2Client {
 	client := &B2Client{
 		HTTPClient:       http.DefaultClient,
 		ApplicationKeyID: applicationKeyID,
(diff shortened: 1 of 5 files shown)


Technifocal:
Out of interest, why was this pull request closed?

Mossop:
Largely accidentally. I was making some changes to my fork and deleted the branch this PR was based on not realising it would close this.

Technifocal:
If you wouldn't mind, I'd love to see this pull request opened again (or recreated).

This is a feature I'd love to see merged into the upstream copy of Duplicacy, rather than running a fork.

fracai:
Figured I'd mention it here that I have updated this PR for the current code base with #589

PR #545, opened 2019-02-21
Only log changed files when backing up with -hash.

Mossop:
This fixes the issue described at https://forum.duplicacy.com/t/why-do-hash-backups-show-every-single-file-being-uploaded/1808.

It works by checking if the file actually changed and then adding it to either preservedEntries or the new list changedEntries. The latter is purely used for display purposes. I suspect there is a way to just modify modifiedEntries an uploadedEntries instead but I'm not sure I understand the snapshot format well enough for that.

I'm not entirely happy with adding the RemoteEntry property to Entry but it seemed the easiest way to do this.

### DIFF
--- a/src/duplicacy_backupmanager.go
+++ b/src/duplicacy_backupmanager.go
@@ -320,7 +320,7 @@ func (manager *BackupManager) Backup(top string, quickMode bool, threads int, ta
 				j++
 			} else if local.Path == remote.Path {
 				local.RemoteEntry = remote
-				if !quickMode && local.IsSameAs(remote) {
+				if quickMode && local.IsSameAs(remote) {
 					local.Hash = remote.Hash
 					local.StartChunk = remote.StartChunk
 					local.StartOffset = remote.StartOffset
--- a/src/duplicacy_backupmanager.go
+++ b/src/duplicacy_backupmanager.go
@@ -198,7 +198,7 @@ func (manager *BackupManager) Backup(top string, quickMode bool, threads int, ta
 		return true
 	}
 
-	// This cache contains all chunks referenced by last snasphot. Any other chunks will lead to a call to
+	// This cache contains all chunks referenced by last snapshot. Any other chunks will lead to a call to
 	// UploadChunk.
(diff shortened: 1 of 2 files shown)


src/duplicacy_backupmanager.go
```diff
@@ -544,6 +547,13 @@ func (manager *BackupManager) Backup(top string, quickMode bool, threads int, ta
 				entry.Size = fileSize
 				uploadedEntries = append(uploadedEntries, entry)
 
+				// Did the file actually change?
+				if entry.RemoteEntry != nil && entry.RemoteEntry.Hash == entry.Hash && entry.RemoteEntry.Size == entry.Size {
```
gilbertchen: This is wrong.  Note that the local hash was just copied from the remote hash previously (on line 324) so the two hashes are always equal.

gilbertchen:
This commit is incorrect.  Please see the review comment as well as my post on the forum:


> -hash doesn’t mean file comparison by hashes. Rather, it means that all files are treated as new files. So Duplicacy would just blindly split and upload every file. In this sense those log messages and stats are correct.
> 
> Note that to do hash comparison you need to scan local files to get local hashes. In the worst case, you’ll need to scan the file once, only to find out that the hash has changed so you’ll need another pass to split the file into chunks. In the current implement only one pass is needed.

src/duplicacy_backupmanager.go
Mossop: So from what I've understood (entirely possible I've missed something else) at this point entry.Size and entry.Hash have been updated to the file's current hash and size, as calculated by the chunkmaker. The entry's RemoteEntry however is the entry as found from the previous snapshot, and so contains the file's hash and size as they were in the previous backup. So they can be different if the file has changed, and testing this locally shows that the hashes are indeed different when the file has been changed.

But now you've pointed out that the entry's hash is originally set to the previous hash before this function is called I realise that I don't need the RemoteEntry at all and can just compare entry.Hash to the new hash before overwriting it in line 546. That seems a lot cleaner.

PR #547, opened 2019-02-28
Add option to use a ssh key signed with a certificate to authenticate

philband:
### Intention
For connecting via SSH there is also an option to use a keypair that was signed by a certificate authority trusted by the server. This enables the dynamic creation short-lived access to backup locations. 
For authentication the client then presents the server its keypair and additionally the certificate.

### Changes
Adding a new configuration variable *DUPLICACY_SSH_CERT_FILE* that is optional and points to the certificate that shall be presented along with the keypair. As it is optional, the user workflow does not change and it is completely backwards compatible.

### DIFF
--- a/src/duplicacy_storage.go
+++ b/src/duplicacy_storage.go
@@ -367,7 +367,33 @@ func CreateStorage(preference Preference, resetPassword bool, threads int) (stor
 				}
 			}
 
-			if key != nil {
+			certFile := GetPasswordFromPreference(preference, "ssh_cert_file")
+			var pubKey ssh.PublicKey
+			var certSigner ssh.Signer
+
+			if certFile != "" {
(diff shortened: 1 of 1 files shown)


philband:
@gilbertchen friendly ping, is there a chance for this to get merged in the foreseeable future?

gilbertchen:
Sorry for not responding earlier.  It looks good.  I'll run a test myself and then merge it before next release.

gilbertchen:

gilbertchen:
I made a change to remove `ssh_cert_file`: https://github.com/gilbertchen/duplicacy/commit/ffac83dd80ece81e923e37e338d30866fe36f024.

It looks like ssh automatically finds the signed certificate file by adding the suffix `-cert.pub` to the key file, so I think we can do the same here.

gilbertchen:

https://forum.duplicacy.com/t/cli-release-2-5-0/3448/1

PR #549, opened 2019-03-01
Improve WebDAV performance

Jos635:
This PR does two things:

* Add `io.Copy(ioutil.Discard, resp.Body)`, which is needed for proper connection reuse, as described here: https://awmanoj.github.io/tech/2016/12/16/keep-alive-http-requests-in-golang/
* Use a non-rate limited reader for uploading when there is no rate limit set. This ensures that the file is sent in one chunk instead of multiple. Some WebDAV servers don't like chunked transfer, and take longer to process files sent that way.

The current code performs as follows for me:

duplicacy-chunked

And with this PR, it looks like this:

duplicacy-nonchunked

Please forgive any stupid mistakes I might have made. This is quite literally my first time doing anything in go.

### DIFF
--- a/src/duplicacy_webdavstorage.go
+++ b/src/duplicacy_webdavstorage.go
@@ -128,7 +128,12 @@ func (storage *WebDAVStorage) sendRequest(method string, uri string, depth int, 
 			dataReader = bytes.NewReader(data)
 		} else if method == "PUT" {
 			headers["Content-Type"] = "application/octet-stream"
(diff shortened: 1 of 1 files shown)


gilbertchen:

https://forum.duplicacy.com/t/error-too-many-open-files/3155/33

gilbertchen:
Thank you for the fixes.  This seems to fix https://forum.duplicacy.com/t/error-too-many-open-files/3155 -- I should have merged this much earlier...

gilbertchen:

Issue #539, opened 2019-02-10
Duplicacy init for B2 storage still ask for account ID


icefo:
I noticed that duplicacy still ask for the account ID when you initialize a backup when it should ask for the applicationKeyId. If you supply the applicationKeyId instead of the account id it works. It can create some confusion.

this is related to merged PR in issue #475

PR #554, opened 2019-04-03
Request B2 "Backblaze Account or Application ID"

arikorn:
(rather than "Account ID")

fixes issue #539 (Duplicacy init for B2 storage still ask for account ID)


gilbertchen:
Thank you for this fix.

Issue #309, opened 2017-12-20
Error: The process cannot access the file because another process has locked a portion of the file.


TheBestPessimist:
```7-12-20 13:00:47.695 TRACE PACK_START Packing c_link/Users/crist/AppData/Local/Packages/Microsoft.SkypeApp_kzf8qxf38zg5c/LocalState/DataRv/offline-storage-ecs.data
2017-12-20 13:00:47.695 INFO PACK_END Packed c_link/Users/crist/AppData/Local/Packages/Microsoft.SkypeApp_kzf8qxf38zg5c/LocalState/DataRv/offline-storage-ecs.data (87862)
2017-12-20 13:00:47.695 TRACE PACK_START Packing c_link/Users/crist/AppData/Local/Packages/Microsoft.SkypeApp_kzf8qxf38zg5c/LocalState/DataRv/offline-storage.data
2017-12-20 13:00:47.695 INFO PACK_END Packed c_link/Users/crist/AppData/Local/Packages/Microsoft.SkypeApp_kzf8qxf38zg5c/LocalState/DataRv/offline-storage.data (4096)
2017-12-20 13:00:47.695 TRACE PACK_START Packing c_link/Users/crist/AppData/Local/Packages/Microsoft.SkypeApp_kzf8qxf38zg5c/LocalState/DataRv/offline-storage.data-shm
2017-12-20 13:00:47.695 ERROR CHUNK_MAKER Failed to read 0 bytes: read D:\backup_software_test\duplicacy repo/c_link/Users/crist/AppData/Local/Packages/Microsoft.SkypeApp_kzf8qxf38zg5c/LocalState/DataRv/offline-storage.data-shm: The process cannot access the file because another process has locked a portion of the file.
```
Is it okay that duplicacy just dies when it cannot read a specific file? I thought it skips it and then carries on.
(yes i am backing up my whole C drive. i am doing this because i am trying to make a good  filter file, which i intend to give it to you, if you want to add it, if not in code, at least in wiki -- as a start)

gilbertchen:
The current implementation is like this:  if a file can't be opened at all, it will be skipped;  however, if a file can be opened but can't be read, then Duplicacy will bail out with an error.  It can't simply skip this file because some bytes from this file may have been already put into a chunk and now it is in an inconsistent state.

However, in the case where no bytes have been read when a read error occurs, it may be safe to skip the file.

In your case if you enable the -vss option that particular file may be readable.

[...]

PR #559, opened 2019-05-01
Zero-length and effectively zero-length handling issues

arikorn:
This PR fixes several issues related to zero-length files. 
1. Most notably fixes Issue #309 "Partially locked files" (fix is in file reader so that the "partially locked" file is treated exactly the same as "fully locked" files; the fix further assumes that the "partially locked" file returns an error upon reading the first few bytes, which has been true in all cases on my Windows 10 system.)
2. Fixes inconsistency in how zero-length files are reported. In v2.1.2, fewer total files are reported on incrementals than on initial backups because full backups count zero-length files whereas incrementals didn't.  Also solves: https://forum.duplicacy.com/t/backup-of-zero-byte-files/1376
3. Fixes a potentially fatal bug in which zero-length chunks are created and then associated with non-zero-length files. See https://github.com/gilbertchen/duplicacy/pull/456#issuecomment-471156048

(Sorry, have been sitting on this branch for a month but hadn't had time to organize this description.)

### DIFF
--- a/src/duplicacy_backupmanager.go
+++ b/src/duplicacy_backupmanager.go
@@ -137,6 +137,11 @@ func setEntryContent(entries []*Entry, chunkLengths []int, offset int) {
 			if totalChunkSize+int64(length) == totalFileSize {
 				entries[i].StartChunk = j + 1 + offset
 				entries[i].StartOffset = 0
+				if entries[i].Size == 0 {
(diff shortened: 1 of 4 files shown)


gilbertchen:

https://forum.duplicacy.com/t/backup-of-zero-byte-files/1376/4

PR #562, opened 2019-05-08
Retry on broken pipe in Azure backend

### LINKED ISSUE #507: Regression: v2.1.1 brings back Azure upload chunk / connection timed out errors
In issue #361, I reported problems. These were fixed in Duplicacy `v2.1.0f`. Some of the errors were apparently in the Azure Go communications code, as I understand it.

These issues appear to be back in Duplicacy `v2.1.1`, unfortunately. Please see the following log file:

First:

```
Uploaded chunk 75838 size 3097840, 14.51MB/s 12:35:55 35.3%
Uploaded chunk 75829 size 11451571, 14.51MB/s 12:35:53 35.3%
Uploaded chunk 75839 size 1345790, 14.51MB/s 12:35:52 35.3%
Failed to upload the chunk 5fdeabae43817238cba56cc2f1c2a5b4248f70b20b9a20de1af84770cb1789b8: Put <path>: write tcp 192.168.1.125:59208->52.191.176.36:443: write: broken pipe
Incomplete snapshot saved to <path>/.duplicacy/incomplete
```

After a while, upon restarting:

```
Uploaded chunk 140240 size 4130921, 19.64MB/s 04:50:35 66.3%
Uploaded chunk 140243 size 4148569, 19.64MB/s 04:50:35 66.3%
Failed to upload the chunk 3d110de3c9ddf3ebff1e1dd70e21af6fd1db2e4f06bdc4f0b0b8cb7cb43a0e86: Put <path>: write tcp 192.168.1.125:61306->52.191.176.36:443: write: broken pipe
Incomplete snapshot saved to <path>/.duplicacy/incomplete
```

And then, yet again:

```
Uploaded chunk 886 size 5205746, 10.75MB/s 08:43:07 1.3%
Uploaded chunk 889 size 2481916, 10.75MB/s 08:42:51 1.3%
Failed to upload the chunk bc76b25b7ad83e38a49afa46840ff6e163b2ee243439d7dd5612f4c25a3974b4: Put <path>: net/http: HTTP/1.x transport connection broken: write tcp 192.168.1.125:49478->52.191.176.36:443: write: broken pipe
Uploaded chunk 890 size 2892104, 10.76MB/s 08:42:32 1.3%
Uploaded chunk 891 size 2692402, 10.74MB/s 08:43:28 1.3%
Incomplete snapshot saved to <path>/.duplicacy/incomplete
```

I think that `broken pipe` should be an error that is retried, as was in `v2.1.0f`. Can this be taken care of?

gilbertchen:
Azure sometimes disconnects the connection randomly when uploading files.  The returned error is 'broken pipe' but this error is wrapped deep in multiple levels of errors so we have to check the error string instead.

This fixed #507.  I mentioned there that I planned to add a retry mechanism to the chunk uploader but eventually I decided to add it to the Azure backend only.  This is because some backends (like sftp and b2) already have their own retry mechanisms so adding another retry layer seems unnecessary.

### DIFF
--- a/src/duplicacy_azurestorage.go
+++ b/src/duplicacy_azurestorage.go
@@ -166,9 +166,21 @@ func (storage *AzureStorage) DownloadFile(threadIndex int, filePath string, chun
 
 // UploadFile writes 'content' to the file at 'filePath'.
 func (storage *AzureStorage) UploadFile(threadIndex int, filePath string, content []byte) (err error) {
-	reader := CreateRateLimitedReader(content, storage.UploadRateLimit/len(storage.containers))
-	blob := storage.containers[threadIndex].GetBlobReference(filePath)
-	return blob.CreateBlockBlobFromReader(reader, nil)
+
+	tries := 0
+
+	for {
+		reader := CreateRateLimitedReader(content, storage.UploadRateLimit/len(storage.containers))
+		blob := storage.containers[threadIndex].GetBlobReference(filePath)
+		err = blob.CreateBlockBlobFromReader(reader, nil)
+
+		if err == nil || !strings.Contains(err.Error(), "write: broken pipe") || tries >= 3 {
+			return err
+		}
+
+		LOG_INFO("AZURE_RETRY", "Connection unexpectedly terminated: %v; retrying", err)
+		tries++
+	}
 
 }
 


TheBestPessimist:
Imo having a global way of setting "retry this many times" (where i would put something like `200_000`) would be really useful. Right now it seems only backblaze got an env variable which makes me sad. 

I would prefer either 
- a config setting (same number of retries for everything?)
- a command parameter (`duplicacy -d -log -retries 200000 backup`)

PR #563, opened 2019-05-20
Add "no follow links" preference

presto8:
I am not sure the history or design philosophy behind the current symlink behavior. However, for my application, I never want symlinks to be followed. In case it is helpful to others, I created this commit which implements a "no_follow_links" preference setting. For now, it is simply true or false. True means to never follow symlinks. False means to use the default symlink behavior (follow top-level links only I think). This could potentially be expanded to be a more generic preference, such as "follow_symlinks" with values of "never", "always", "top-level", etc.

Add a new preference "no_follow_links" which can be set in
.duplicacy/preferences to override the default symlink behavior.

If the preference is not set or is set to false, then the symlink
behavior is the same as before this commit.

If the preference is set to true, then symlinks are never followed. The
symlink is stored as a regular file in the backup.

### DIFF
--- a/duplicacy/duplicacy_main.go
+++ b/duplicacy/duplicacy_main.go
@@ -532,6 +532,11 @@ func setPreference(context *cli.Context) {
 		newPreference.DoNotSavePassword = triBool.IsTrue()
 	}
 
+	triBool = context.Generic("no-follow-links").(*TriBool)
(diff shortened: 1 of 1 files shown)


presto8:
Oops, forgot to include a file.... I will resubmit.

PR #566, opened 2019-05-24
Update the issue template

TheBestPessimist:
Use the new template format and ask people to use the forum **more thoroughly**.


TheBestPessimist:
Thanks for merging :)

PR #571, opened 2019-07-11
Added noagent flag to add and init

Foosec:
I ran into an issue with it using ssh-agent causing too many authentication failures errors for sftp.
Some random styling changes and empty line deletes are caused by visual studio code, if its a problem ill gladly edit it somewhere else to leave it as it is.

### DIFF
--- a/duplicacy/duplicacy_main.go
+++ b/duplicacy/duplicacy_main.go
@@ -24,7 +24,7 @@ import (
 
 	"io/ioutil"
 
-	"github.com/gilbertchen/duplicacy/src"
+	duplicacy "github.com/gilbertchen/duplicacy/src"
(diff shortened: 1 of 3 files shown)


TheBestPessimist:
@Foosec did you figure out what was the problem with the ssh-agent?

Foosec:
@TheBestPessimist Its because it tries to use ssh-agent before a user specified key it tries too many wrong ones before getting to the correct one, causing the server to disconnect due to too many failed attempts. I closed it because i decided i should have a second look into how i solved it, it was a quick first fix and i forgot to check if ssh is used anywhere else besides init and add. I will create a new pull request once i am satisfied with the changes.

PR #572, opened 2019-07-22
Added identities-only global flag, similar to ssh_config's IdentitiesOnly

Foosec:
It prevents the usage of ssh-agent for ssh authentication possibly causing the server to disconnect with "Too many authentication failures".
Random styling "fixes" courtesy of Visual studio code... If these present an issue ill fix them by hand.


Foosec:
I kindoff forgot about this; feel free to close the PR

PR #575, opened 2019-09-20
Implement RSA encryption

gilbertchen:
This is to support public key encryption in the backup operation.  You can use
the `-key` option to supply the public key to the backup command, and then the
same option to supply the private key when restoring a previous revision.

The storage must be encrypted for this to work.


gilbertchen:

https://forum.duplicacy.com/t/new-feature-rsa-encryption/2662/1

gilbertchen:

https://forum.duplicacy.com/t/new-feature-rsa-encryption/2662/1

PR #593, opened 2020-04-08
Call ReadAll() on the body io.ReadCloser to allow the http keepalive connection to be reused.

freaksdotcom:
This should improve performance on file downloads.


gilbertchen:
Thanks for the fix!

gilbertchen:

https://forum.duplicacy.com/t/cli-release-2-5-0/3448/1

PR #594, opened 2020-05-05
Call Authenticate() before using Swift storage

alecuyer:
Failing to call Authenticate() before using the swift connection will
cause a panic in recent versions of the swift library.


gilbertchen:
Thank you for this fix!

gilbertchen:

Issue #205, opened 2017-09-20
Restore should not come to a full stop when a corrupt chunk is encountered.


macdanny:
This is partly a continuation of the "Duplicacy keeps stalling" issue on the duplicacy web site. I am not allowed to add more comments there because I have "reached my post limit".

---

I switched to the s3c provider, cranked up the threads to 128, and let it run. It got about 600 chunks in and then crashed:

```
2017-09-20 12:50:45.539 ERROR UPLOAD_CHUNK Failed to decrypt the chunk ca9c51caa3e5cae3e48d72b65cd80618b7e27537b51732c54e884efcce9adc88: The storage doesn't seem to be encrypted
goroutine 134 [running]:
runtime/debug.Stack(0x0, 0x0, 0xc42a62d6b8)
        /usr/local/go/src/runtime/debug/stack.go:24 +0x79
runtime/debug.PrintStack()
        /usr/local/go/src/runtime/debug/stack.go:16 +0x22
github.com/gilbertchen/duplicacy/src.CatchLogException()
        /Users/chgang/zincbox/go/src/github.com/gilbertchen/duplicacy/src/duplicacy_log.go:177 +0x13d
panic(0xc6d3a0, 0xc42834a900)
        /usr/local/go/src/runtime/panic.go:489 +0x2cf
github.com/gilbertchen/duplicacy/src.logf(0x2, 0xd1868f, 0xc, 0xd2c6f6, 0x22, 0xc42a62dd40, 0x2, 0x2)
        /Users/chgang/zincbox/go/src/github.com/gilbertchen/duplicacy/src/duplicacy_log.go:160 +0x1bc
github.com/gilbertchen/duplicacy/src.LOG_ERROR(0xd1868f, 0xc, 0xd2c6f6, 0x22, 0xc42a62dd40, 0x2, 0x2)
        /Users/chgang/zincbox/go/src/github.com/gilbertchen/duplicacy/src/duplicacy_log.go:98 +0x73
github.com/gilbertchen/duplicacy/src.(*ChunkDownloader).Download(0xc42876cf00, 0x47, 0x0, 0x27d, 0xc42abb77e0, 0x20, 0x331d30, 0x1, 0xc4265e4e01)
        /Users/chgang/zincbox/go/src/github.com/gilbertchen/duplicacy/src/duplicacy_chunkdownloader.go:336 +0x547
[...]

PR #595, opened 2020-05-06
Adds -persist option to check and restore commands to continue despite errors

### LINKED ISSUE #157: Enhancement - Backup recovery with missing chunks
As the backup grows huge, the number of chunks are bound to increase and can run into 100 thousands. A few (or hundred) missing chunks shouldn't render the backup useless. Please allow recovering from backups with missing chunks.

I believe the working is this way
1. The initial chunk to be read is stored somewhere in the snapshot file
2. Each chunk has a pointer to the next chunk

A restore should restore all files which it can restore fully and Warn that there can 1 or many files that couldn't be restored due to missing pieces.

Similarly, it shouldn't prevent from taking additional backups. It should warn that there are missing chunks and hence a full scan should be done (-hash). The chunks that are having parts of corrupted files should be put into fossil and new parts should be created.

Hope you give this a thought and implement it in a wonderful way.

### LINKED ISSUE #596: duplicacy restore fails with -hash if existing file present with correct hash
Duplicacy `restore` will fail if the `-hash` parameter is specified without `-overwrite`, even if an existing file with the correct hash is present and no actual file overwrite will occur. The expected behaviour should be that failure will only occur if a file overwrite is actually necessary. Enabling `-overwrite` is not a suitable workaround, as users may wish to restore any missing files while being sure that no existing files are overwritten.

**Minimal working example:**
Perform backup: `duplicacy backup`
```
Storage set to /data/duplicacy_test/backup
No previous backup found
Indexing C:\data\duplicacy_test\to_backup
Parsing filter file \\?\C:\data\duplicacy_test\to_backup\.duplicacy\filters
Loaded 0 include/exclude pattern(s)
Packed testfile1(816)
...
Backup for C:\data\duplicacy_test\to_backup at revision 1 completed
```

Do in-place restore with `-hash`:  `duplicacy restore -r 1 -hash`
```
Storage set to /data/duplicacy_test/backup
Loaded 0 include/exclude pattern(s)
Parsing filter file \\?\C:\data\duplicacy_test\to_backup\.duplicacy\filters
Loaded 0 include/exclude pattern(s)
Restoring C:\data\duplicacy_test\to_backup to revision 1
File testfile1 already exists.  Please specify the -overwrite option to continue
```
Note that `testfile` is unchanged, and does not need to be overwritten, as shown here:
`duplicacy restore -r 1 -hash -overwrite -stats`
```
Storage set to /data/duplicacy_test/backup
Loaded 0 include/exclude pattern(s)
Parsing filter file \\?\C:\data\duplicacy_test\to_backup\.duplicacy\filters
Loaded 0 include/exclude pattern(s)
Restoring C:\data\duplicacy_test\to_backup to revision 1
Restored C:\data\duplicacy_test\to_backup to revision 1
Files: 242 total, 416K bytes
Downloaded 0 file, 0 bytes, 0 chunks
Total running time: 00:00:03
```
Note 0 files downloaded. `-verbose` mode (not shown) indicates all files are "unchanged (by hash)".

This behaviour with `-hash` is inconsistent with the behaviour without `-hash`:
`duplicacy restore -r 1 -stats`
```
Storage set to /data/duplicacy_test/backup
Loaded 0 include/exclude pattern(s)
Parsing filter file \\?\C:\data\duplicacy_test\to_backup\.duplicacy\filters
Loaded 0 include/exclude pattern(s)
Restoring C:\data\duplicacy_test\to_backup to revision 1
Restored C:\data\duplicacy_test\to_backup to revision 1
Files: 0 total, 0 bytes
Downloaded 0 file, 0 bytes, 0 chunks
Total running time: 00:00:01
```
Note that this completes successfully without the `-overwrite` option.

twlee79:
Addresses #157  and #205 and fixes #596

This PR adds a `-persist` option to the `check` and `restore` modes of duplicacy. The function of this option is to allow these processes to continue despite encountering errors when downloading or decrypting file chunks. This allows `check` or `restore` to continue until all chunks/files in the revision are processed. In `check`, this means all chunks with errors (`-chunks` mode) or all files with errors (`-files` mode) are reported. In `restore`, this means that duplicacy will attempt to restore all files rather than failing on first error. Only errors in file chunks can be recovered from; errors in metadata chunks will still cause fatal errors. 

In my view this PR increases the robustness of backups as they can be restored as far as possible despite corrupt/missing file chunks. Assuming only a small number of chunks are missing/corrupted, this reduces a potentially a critical failure (inability to restore entire backup) to a  lesser failure (only several files corrupted). It is worth noting that this increased robustness only applies to loss of file chunks. Loss of a metadata chunk is likely to still lead to loss of an entire backup. Therefore, this PR can be considered to mostly fix #157  and #205.

Duplicacy function without the `-persist` option is largely the same as the current version (the difference is for behaviour with `restore -hash` without `-overwrite` to fix #596, and additional statistics with `restore -stats`).

**Internal working:**
The `-persist` option is propagated to the `Restore()` and `CheckSnapshops()` through the use of a new parameter `allowFailures` (default `allowFailures=false`; `allowFailures=true` if `-persist` enabled). 

*Effect on chunk downloader*
The `allowFailures` parameter is propagated to `CreateChunkDownloader()` by `Restore()` and `CheckSnapshops()`. With `allowFailures==true`, the chunk downloader will produce warnings on chunk download/decrypt failures rather than fatal errors (following the standard number of retries), with chunks marked with `isBroken=true`. Therefore, with `-persist`, missing/corrupted file chunks will only affect the files encoded by these chunks.

*Effect on `CheckSnapshots()`*
The `allowFailures` parameter is propagated to `CreateChunkDownloader()`. In `CheckSnapshots()` and in `-chunks` mode, file chunk failures will be output as warnings; verification will continue until all chunks processed (so warnings will be produced for all affected chunks). In `-files` mode, corrupted files will be reported as well as total number of files affected. 

*Effect on `Restore()`*
The `allowFailures` parameter is propagated to `CreateChunkDownloader()` and `RestoreFile()`. In `RestoreFile()`, `allowFailures=true` will cause failures to reduce to warnings rather than errors, including existing file overwrite with `overwrite==false` and restored file hash failures. Chunks with `isBroken==true` are not used for reassembling the file and the resulting file will be corrupt and result in file hash errors. `RestoreFile()` is altered to return `bool, Error` (rather than just `bool`). The `bool` option indicates if the file was restored or not; the `error` option indicates if an error was encountered.  The three return possibilities are `true, nil`: file restored; `false, nil`: file skipped (already present with correct hash); and `false, error`: file could not be restored. With this change, `Restore()` will track the `downloadedFiles`, `skippedFiles` and `failedFiles`, reporting totals for all if `-stats` requested. 

*Restore differences without `-persist`*
Without `-overwrite` and with `-hash`, duplicacy would previously fail if file is present even if file has same hash (see #596). This behaviour has been changed by shifting the exiting file hash check further up in `RestoreFile()` and having `RestoreFile()` to signal a skipped file if an identical existing file is present (fixing #596); now `-overwrite` off will only cause failures if the file exists but has a different hash. 

[This changes also works well with `restore -hash -persist` (`overwrite=false`) as it allows a user to restore all missing files without any overwrites; any files with different hashes that could not be restored due to existing files are reported.]
 
In addition, restore will report more statistics (`skippedFiles` and `failedFiles`) regardless of `-persist`.

*Additional unit tests*
Additional unit tests for `allowFailures=true` have been added to `duplicacy_backupmanager_test.go`: `TestBackupManagerPersist`

### DIFF
--- a/duplicacy/duplicacy_main.go
+++ b/duplicacy/duplicacy_main.go
@@ -794,6 +794,7 @@ func restoreRepository(context *cli.Context) {
 	setOwner := !context.Bool("ignore-owner")
 
 	showStatistics := context.Bool("stats")
+	persist := context.Bool("persist")
 
 	var patterns []string
 	for _, pattern := range context.Args() {
@@ -824,7 +825,7 @@ func restoreRepository(context *cli.Context) {
 	loadRSAPrivateKey(context.String("key"), preference, backupManager, false)
 
 	backupManager.SetupSnapshotCache(preference.Name)
-	backupManager.Restore(repository, revision, true, quickMode, threads, overwrite, deleteMode, setOwner, showStatistics, patterns)
+	backupManager.Restore(repository, revision, true, quickMode, threads, overwrite, deleteMode, setOwner, showStatistics, patterns, persist)
 
 	runScript(context, preference.Name, "post")
 }
@@ -934,9 +935,10 @@ func checkSnapshots(context *cli.Context) {
 	checkChunks := context.Bool("chunks")
 	searchFossils := context.Bool("fossils")
 	resurrect := context.Bool("resurrect")
+	persist := context.Bool("persist")
 
 	backupManager.SetupSnapshotCache(preference.Name)
-	backupManager.SnapshotManager.CheckSnapshots(id, revisions, tag, showStatistics, showTabular, checkFiles, checkChunks, searchFossils, resurrect, threads)
+	backupManager.SnapshotManager.CheckSnapshots(id, revisions, tag, showStatistics, showTabular, checkFiles, checkChunks, searchFossils, resurrect, threads, persist)
 
 	runScript(context, preference.Name, "post")
 }
@@ -1510,6 +1512,10 @@ func main() {
 					Usage:    "the RSA private key to decrypt file chunks",
 					Argument: "<private key>",
 				},
+				cli.BoolFlag{
+					Name:  "persist",
+					Usage: "continue processing despite chunk errors or existing files (without -overwrite), reporting any affected files",
+				},
 			},
 			Usage:     "Restore the repository to a previously saved snapshot",
 			ArgsUsage: "[--] [pattern] ...",
@@ -1626,6 +1632,10 @@ func main() {
 					Value:    1,
 					Usage:    "number of threads used to verify chunks",
 					Argument: "<n>",
+				},
+				cli.BoolFlag{
+					Name:  "persist",
+					Usage: "continue processing despite chunk errors, reporting any affected (corrupted) files",
 				},
 			},
 			Usage:     "Check the integrity of snapshots",
--- a/src/duplicacy_backupmanager.go
+++ b/src/duplicacy_backupmanager.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -741,7 +742,7 @@ func (manager *BackupManager) Backup(top string, quickMode bool, threads int, ta
 // the same as 'top'.  'quickMode' will bypass files with unchanged sizes and timestamps.  'deleteMode' will
 // remove local files that don't exist in the snapshot. 'patterns' is used to include/exclude certain files.
 func (manager *BackupManager) Restore(top string, revision int, inPlace bool, quickMode bool, threads int, overwrite bool,
-	deleteMode bool, setOwner bool, showStatistics bool, patterns []string) bool {
+	deleteMode bool, setOwner bool, showStatistics bool, patterns []string, allowFailures bool) bool {
 
 	startTime := time.Now().Unix()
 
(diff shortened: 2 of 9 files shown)


gilbertchen:

https://forum.duplicacy.com/t/improving-robustness-of-duplicacy-to-missing-corrupt-chunks-with-a-secondary-snapshot-table/3627/1

gilbertchen:

https://forum.duplicacy.com/t/add-option-for-restore-command-to-pass-filters-file/3916/1

gboudreau:
@gilbertchen Any particular reason this has not been merged yet? Would you maybe prefer this to be handled differently, or maybe merged separately?
I was about to look into the code, to see how I could create a PR that would allow `-check` to report all invalid chunks, instead of dying at the first invalid chunk found. But this PR goes further.

gilbertchen:
Sorry I wasn't really paying attention to this PR.  This is exactly what we need to fix issues caused by the sftp bug.

@twlee79 Thank you for the contribution!  This is a high quality PR that includes excellent tests and docs.  There are a few minor changes that I will make after the merge.

gilbertchen:

PR #599, opened 2020-06-07
Initial Glacier Support

juju2013:
Hi,

Here's my changes to make backup directly from/to Glacier storage via the S3CStorage driver.


edo1:
@gilbertchen any chance to see glacier support integrated?

PR #606, opened 2020-09-03
Implement Erasure Coding

gilbertchen:
A new option `-erasure-coding` is added to the `init` and `add` commands to provide the ability to recover from corrupted chunks.

Example use:

```
duplicacy init -erasure-coding 5:2 repository_id storage_url
```

When restoring from a storage with erasure coding enabled, the following log messages show corrupted chunks are being repaired:

```
  Restoring /private/tmp/duplicacy_test/repository to revision 1
  Recovering a 1824550 byte chunk from 364910 byte shards: ***--**
  Downloaded chunk 1 size 1817347, 1.73MB/s 00:00:11 9.0%
  Recovering a 6617382 byte chunk from 1323477 byte shards: **--***
  Downloaded chunk 2 size 6591322, 8.02MB/s 00:00:02 42.0%
  Recovering a 5136934 byte chunk from 1027387 byte shards: --*****
  Downloaded chunk 3 size 5116593, 12.90MB/s 00:00:01 67.6%
  Recovering a 2515494 byte chunk from 503099 byte shards: -*****-
  Downloaded chunk 4 size 2505558, 15.29MB/s 00:00:01 80.1%
  Recovering a 3984934 byte chunk from 796987 byte shards: --*****
  Downloaded chunk 5 size 3969180, 19.07MB/s 00:00:01 100.0%
  Downloaded file1 (20000000)
```

### DIFF
--- a/duplicacy/duplicacy_main.go
+++ b/duplicacy/duplicacy_main.go
@@ -458,8 +458,26 @@ func configRepository(context *cli.Context, init bool) {
 		if iterations == 0 {
 			iterations = duplicacy.CONFIG_DEFAULT_ITERATIONS
 		}
+
+		dataShards := 0
+		parityShards := 0
+		shards := context.String("erasure-coding")
+		if shards != "" {
+			shardsRegex := regexp.MustCompile(`^([0-9]+):([0-9]+)$`)
+			matched := shardsRegex.FindStringSubmatch(shards)
+			if matched == nil {
+				duplicacy.LOG_ERROR("STORAGE_ERASURECODE", "Invalid erasure coding parameters: %s", shards)
+			} else {
+				dataShards, _ = strconv.Atoi(matched[1])
+				parityShards, _ = strconv.Atoi(matched[2])
+				if dataShards == 0 || dataShards > 256 || parityShards == 0 || parityShards > dataShards {
+					duplicacy.LOG_ERROR("STORAGE_ERASURECODE", "Invalid erasure coding parameters: %s", shards)
+				}
+			}
+		}
+
 		duplicacy.ConfigStorage(storage, iterations, compressionLevel, averageChunkSize, maximumChunkSize,
-			minimumChunkSize, storagePassword, otherConfig, bitCopy, context.String("key"))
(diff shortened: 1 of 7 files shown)


gilbertchen:

https://forum.duplicacy.com/t/new-feature-erasure-coding/4168/1

duplicacy/duplicacy_main.go
```diff
@@ -1403,6 +1421,11 @@ func main() {
 					Usage:    "the RSA public key to encrypt file chunks",
 					Argument: "<public key>",
 				},
+				cli.StringFlag{
+					Name:     "erasure-coding",
+					Usage:    "enable erasure coding to protect against storage corruption",
```
TheBestPessimist, line 1426: About naming: 

From https://forum.duplicacy.com/t/new-feature-erasure-coding/4168/2?u=thebestpessimist i understand that this only protects the same chunk from corruption, and not the whole storage (ie. like raid5/6 does).

I think the Usage should be reworded to something like `enable erasure coding to protect against intra-chunk corruption`, as this is not "storage level protection", just "chunk-level protection".


What would happen if in the future you would also add something like storage-level protection via something similar to raid5/6, which would do parity for any chunk?

src/duplicacy_chunk.go
```diff
@@ -68,11 +70,13 @@ type Chunk struct {
 }
 
 // Magic word to identify a duplicacy format encrypted file, plus a version number.
-var ENCRYPTION_HEADER = "duplicacy\000"
+var ENCRYPTION_BANNER = "duplicacy\000"
 
 // RSA encrypted chunks start with "duplicacy\002"
 var ENCRYPTION_VERSION_RSA byte = 2
 
+var ERASURE_CODING_BANNER = "duplicacy\003"
```
TheBestPessimist, line 78: I could be missunderstanding stuff but
`RSA encrypted chunks start with "duplicacy\002"`
`ERASURE_CODING_BANNER = "duplicacy\003"`
does this mean that we cannot have rsa and erasure at the same time?

TheBestPessimist: If we can have all of those, then the flag `ENCRYPTION_BANNER ` could be better renamed to `FEATURE_LEVEL`.

PR #614, opened 2020-12-09
Dropbox storage retrying with generic RetryableStorage interface

overhacked:
**Retries implemented for Dropbox `UploadFile()` and `DeleteFile()`**

Also standardizes an interface to configure retrying of storage operations for other providers. I have work-in-progress code to submit that converts the other storage providers to the generic `RetryableStorage` interface, but I didn't want to bloat this PR. I can open a second PR before this is merged, if you'd like.

The test code attempts to exceed rate limits for any `RetryableStorage` by firing off 50 concurrent `UploadFile()` operations, then waiting for all to succeed. This level of concurrency may need to be increased to trigger rate limiting on any excessively generous storage providers.

Adds a dependency on github.com/cenkalti/backoff: used by `RetryableStorage` implementation to provide exponential backoff for storage operation retries.

A version change in `dep` on my machine caused the formatting of `Gopkg.lock` to change significantly. I manually reformatted so that only the non-whitespace changes are part of this commit (to keep the PR shorter).

Before merging, I'd argue that `dep ensure` should be run with the latest or project-adopted version of `dep` so that my manual formatting is reverted to `dep` default behavior.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -32,6 +32,12 @@
   version = "v1.0.0"
 
 [[projects]]
+  name = "github.com/cenkalti/backoff"
+  packages = ["."]
+  revision = "5267b6dd4d2666b980a911bf235efa276222cbe2"
+  version = "v2.2.1"
+
+[[projects]]
   name = "github.com/dgrijalva/jwt-go"
   packages = ["."]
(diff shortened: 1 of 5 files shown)


overhacked:
I see that this was addressed in https://github.com/gilbertchen/go-dropbox/pull/2, so closing this PR.

PR #626, opened 2021-12-17
Swift v2

markfeit:
This updates Swift storage to use the latest version of NCW's Swift package, eliminating the need to do anything special to get the old version when compiling.

I've done some basic testing against Memset (thanks for the suggestion) and it seems to work.


gilbertchen:
Thank you for the fix!

gilbertchen:

https://forum.duplicacy.com/t/unable-to-build-from-source/6232/5

PR #627, opened 2021-12-28
Clean up formatting and warnings

freaksdotcom:
Perform some maintenance on the source files and clean up a potential memory leak.


gilbertchen:
I won't accept format changes.  Besides, a lot of files have been changed since the backup engine rewrite for memory optimization.  But, can you submit the `time.Tick` change as a separate PR?  That looks necessary.

PR #629, opened 2022-05-31
add bacula / bareos

### LINKED ISSUE #620: Comparison to bacula
Could you add bacula to the comparison table?

RubenKelevra:
fixes #620


gilbertchen:
Sorry I won't add bacula because it is quite different from restic/borg/duplicacy.

PR #628, opened 2022-05-07
DRAFT: Allows to mount backups using FUSE

SmArtKar:
You can mount backups using go-fuse by using `duplicacy fusemount <mount folder name> <revision ID>`
Also apparently master is broken right now so I branched off the last working release commit i found.
TODO:
 - unmount

PR #636, opened 2022-10-20
Live backup preview (mount)

davidrios:
This work is inspired by [!628](https://github.com/gilbertchen/duplicacy/pull/628), but it's a more complete implementation.

I'm using [cgofuse](https://github.com/winfsp/cgofuse) so it works on all platforms, including Windows.

The command is just `duplicacy mount POINT`, no revision needed. When you run it for the first time, it'll load the list of revisions and create a base tree organizing them by date using the format `/%YYYY/%MM/%DD/%HH%mm.%REV`. For instance, let's say you have this output from `duplicacy list`:

```
$ duplicacy list
Snapshot Users revision 1 created at 2022-10-19 20:09 -hash
Snapshot Users revision 2 created at 2022-10-19 21:38
Snapshot Users revision 3 created at 2022-10-20 10:38
Snapshot Users revision 4 created at 2022-11-14 13:44
```

The following base tree will be created:

```
2022/
    10/
        19/
            2009.1/
            2138.2/
        20/
            1038.3/
    11/
        14/
            1344.4/
```

There're a couple of reasons for that:

- If you have thousands of revisions, they would make a huge single list
- When you open a folder on Windows explorer, it'll try to list the contents of the next level, which would be bad if all revisions were on the same level
- The revision number is appended to avoid collision for revisions created at the same `HH:mm`

After the initial structure is created, the program will lazily load and create the file structure when you try to list the contents of (or access a file in) one of revision folders for the first time, then it'll be cached until program end.

Folders and files will display their saved attributes, with the caveat that everything that's not a folder is shown as a regular file.

File reading is efficiently implemented by only downloading the chunks needed for the specific OS read request and caching them. The cache is a 2Q LRU cache by [golang-lru](https://pkg.go.dev/github.com/hashicorp/golang-lru#TwoQueueCache) with size 20 and keyed by the chunk hash.

I've tested using a repository with these characteristics:

- 10.2 GB
- 70151 Files
- 10467 Folders
- 4 revisions
- sftp storage on a remote server

It takes a couple of seconds to create the base tree and a few seconds for every revision dir that is loaded for the first time, but otherwise everything works as expected.

### DIFF
--- a/duplicacy/duplicacy_main.go
+++ b/duplicacy/duplicacy_main.go
@@ -24,7 +24,7 @@ import (
 
 	"io/ioutil"
 
-	"github.com/gilbertchen/duplicacy/src"
+	duplicacy "github.com/gilbertchen/duplicacy/src"
 )
 
 const (
@@ -1022,7 +1022,6 @@ func printFile(context *cli.Context) {
 		snapshotID = context.String("id")
 	}
 
-
 	backupManager := duplicacy.CreateBackupManager(preference.SnapshotID, storage, repository, password, "", "", false)
 	duplicacy.SavePassword(*preference, "password", password)
 
@@ -1269,7 +1268,7 @@ func copySnapshots(context *cli.Context) {
 	destinationStorage.SetRateLimits(0, context.Int("upload-limit-rate"))
 
 	destinationManager := duplicacy.CreateBackupManager(destination.SnapshotID, destinationStorage, repository,
-		                                                  destinationPassword, "", "", false)
+		destinationPassword, "", "", false)
 	duplicacy.SavePassword(*destination, "password", destinationPassword)
 	destinationManager.SetupSnapshotCache(destination.Name)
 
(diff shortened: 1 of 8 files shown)


sevimo123:
I haven't tried it yet, but it sounds great! Couple of questions here:
- when/how cached revision list is refreshed (e.g. you did mount, then run another backup, now you want to see new revision in the mount)?
- Does the cache use memory/disk? Is there any way to limit cache usage (e.g. ala rclone vfs parameters), as otherwise it would be possible to run out of memory/disk during long-running mounts.

We'll need to stress test it with some storages with thousands of revisions to see how it scales up. Cheers!

sevimo123:
> The design choices were made on the assumption that the command will be used when needed and only to browse a few revisions. I guess it could be tweaked to support running permanently as a service, if that's a desired use case.

I tend to run my mounts more or less permanently, so that I can have access from anything (scripts etc) without running anything special first. But for this case I don't think it is really necessary as primary use would be to restore, which may never (or very infrequently) happen. I think for the most cases it is perfectly fine to re-mount for refresh.

Though usecase might be just the ability to check that new revisions are getting created by simply looking at filesystem, without running specific duplicacy commands. Probably not too important ;)

davidrios:
To me that sounds strange, but I can definitely understand the appeal. I guess if the feature is there, there's really no reason to not use it 😂

gilbertchen:

https://forum.duplicacy.com/t/mount-duplicacy-snapshot-as-a-virtual-filesystem/1718/18

gilbertchen:

https://forum.duplicacy.com/t/restore-multiple-files-web-ui/5320/21

gilbertchen:

https://forum.duplicacy.com/t/mount-duplicacy-snapshot-as-a-virtual-filesystem/1718/47

cyberguy42:
@davidrios I can't believe this pull request hasn't received more attention. The absence of this functionality was a deal breaker for me when I first tested Duplicacy several years ago. When I saw this pull request about a week ago, I decided to give Duplicacy another go and have not regretted it. Thank you for sharing your efforts with the community!

PR #648, opened 2023-03-21
Upgrade go-dropbox to the latest

northnose:
This commit is the result of `go get github.com/gilbertchen/go-dropbox`, to get https://github.com/gilbertchen/go-dropbox/pull/5


gilbertchen:
Thank you for the fix!

PR #652, opened 2023-08-29
Update readme to reflect current restic capabilities

sammcj:
Update readme comparison table with relation to restic.

- Compression was added in early 2022 https://forum.restic.net/t/compression-support-has-landed-in-master/4997
- Deletion was added in early 2023 https://restic.net/blog/2023-01-12/restic-0.15.0-released/


sammcj:
hmm, I don't really want to sign that big legal document.

PR #656, opened 2023-09-24
Bump to more recent Go version

devusb:
This resolves the following build error when building in nixpkgs:
```
> vendor/github.com/klauspost/compress/huff0/decompress.go:774:21: cannot convert out (variable of type []byte) to type *[256]byte: conversion of slices to array pointers requires go1.17 or later (-lang was set to go1.16; check go.mod)
```


devusb:
Done in https://github.com/gilbertchen/duplicacy/commit/7cc1b4222cb466f9707bd338245f7f22f608e5a7

PR #684, opened 2025-10-25
Change permission before in-place writing or deletion

9-2-1:
Restoring will fail because the file have read-only attribute or read-only permission.

As enabled -overwrite and restore with permission, restoring should overcome this, by set permission to read and write before overwriting.

The same applies to deleting extra files. Grant read and write permission before deleting.

In addition, it's confusing before as failing to delete extra files will be silently ignored. Now it log a warning if here are any errors.


gilbertchen:

https://forum.duplicacy.com/t/access-is-denied-error-while-restoring-into-folder-with-files-having-read-only-attributes/10640/1

